PR #6, opened 2015-09-17
Moved the setting of ulimit from the startup script to the docker file.

britebit:
When I tried to run the docker file on AWS received an error similar to "ulimit: open files: cannot modify limit: Operation not permitted".  It looked similar to the issue at http://stackoverflow.com/questions/24961576/docker-error-while-creating-couchbase-ulimit-open-files-cannot-modify-limit and moving the ulimit command from the startup script to the docker file fixed the problem for me.
I am very new to docker, neo4j and bash scripts, so what I have done might be sub-optimal :-)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,12 @@ RUN sed -i -e "s|.*dbms.pagecache.memory=.*|dbms.pagecache.memory=512M|g" /var/l
     && sed -i -e "s|org.neo4j.server.webadmin.rrdb.location=.*|org.neo4j.server.webadmin.rrdb.location=/tmp/rrd|g" /var/lib/neo4j/conf/neo4j-server.properties \
     && sed -i -e "s|Dneo4j.ext.udc.source=.*|Dneo4j.ext.udc.source=docker|g" /var/lib/neo4j/conf/neo4j-wrapper.conf
 
+RUN if [ -n "$NEO4J_OPEN_FILES" ]; then \
+	ulimit -n $NEO4J_OPEN_FILES > /dev/null \
+else \
+	ulimit -n 40000 > /dev/null; \
+fi
+
 VOLUME /data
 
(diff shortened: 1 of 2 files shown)


jexp:
did you try to run it with --privileged ? 

https://docs.docker.com/reference/run/#runtime-privilege-linux-capabilities-and-lxc-configuration

jexp:
also the enviroment variable is not taken into account during build b/c it is not set, it was a runtime setting

britebit:
Sorry, I forgot one small, tiny detail :$ ... I was tryng to run the container on a single-instance AWS Beanstalk docker envrionment. Running the docker instance locally works just fine. Using the docker file to build an instance in AWS causes the problem I saw.  I understand that this is a beta and not for production\serious use .... I just thought I'd try it :-)

It seems that running the --privileged flag is not recommended (
[Stack Overflow](http://stackoverflow.com/questions/30136047/dockerfile-privileged-flag-for-docker-container-needed-because-of-apache-error))  and I'm not sure it's officially supported in an AWS Beanstalk environment.

You're right about the environment variable not being taken into account.

jbhannah:
Seems like a reasonable compromise might be to put the standard `ulimit -n 4000` in the Dockerfile to handle cases like this, and leave the `if` statement that looks for `$NEO4J_OPEN_FILES` in the startup script, so if a user really needs a different limit they can set it at runtime.

jexp:
@jbhannah yes that sounds good, care to send a PR for it?

jbhannah:
@jexp Done.

benbc:
This has been implemented in #7.

britebit:
Cool, thanks guys.

benbc:
@britebit FYI: I have now take this out altogether and documented the use of docker run --ulimit instead (see 070bd66).

I found that ulimit in the Dockerfile was having no effect because those settings are only per-session.

PR #7, opened 2015-09-22
Move initial ulimit setting to Dockerfile

jbhannah:
Still allows it to be set at runtime by `$NEO4J_OPEN_FILES`. Supersedes neo4j-contrib/docker-neo4j#6.


benbc:
@jbhannah FYI: I have now take this out altogether and documented the use of `docker run --ulimit` instead (see https://github.com/neo4j/docker-neo4j/commit/070bd663e38cac4b655de277e3b74e005ba3e18b).

I found that `ulimit` in the Dockerfile was having no effect because those settings are only per-session.

PR #8, opened 2015-10-05
Add Vagrantfile for building on OS X

jbhannah:
I put this Vagrantfile together to make the build process easier on OS X (my main dev environment for my fork of this repo). All it does is install and start the Docker service on an Ubuntu Vivid base, but from there it only takes a `vagrant ssh` and `cd /vagrant; make` to build the Dockerfiles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+/.vagrant/
 /dev/
 devenv.local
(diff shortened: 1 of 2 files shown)


benbc:
@jbhannah How would you feel about using `docker-machine` for this instead of Vagrant?

benbc:
@jbhannah I've just pushed changes to make the build scripts OSX-compatible. I've tested this with docker-machine running a Virtualbox machine and it works fine. Would you like to try it out?

PR #18, opened 2015-11-07
Updated default host of neo4j

ikwattro:
Docker doesn't run on localhost. I don't know however if `192.168.99.100` is the default docker host ip on all platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The image exposes two ports (`7474` and `7473`) for HTTP and HTTPS access to the
 	    --volume=$HOME/neo4j-data:/data \
 	    neo4j/neo4j
 
-Point your browser at `http://192.168.99.100:7474`. Where here `192.168.99.100` is the ip of the Docker hosts.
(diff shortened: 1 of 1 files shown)


jbhannah:
This only applies on platforms where Docker Machine is necessary (i.e. OS X and Windows); a note, for example, to

> replace `localhost` with your Docker Machine IP address on OS X and Windows

would suffice, instead of making Docker Machine the default assumption.

ikwattro:
@jbhannah cool nice to know. I'll amned with your suggestion.

benbc:
@ikwattro Thanks for raising this. We have had several people run into similar problems on OSX. We're going to try to figure out the best way of avoiding this. (The trouble is that it's not just this `http://localhost:7474` bit that catches people out -- there is stuff with mounting volumes that differs on OSX as well.)

benbc:
I've added some help text in https://github.com/neo4j/docker-neo4j/commit/24259da41e1c6ca88f5e6cdd44146b5d810c156c.

PR #22, opened 2016-01-04
Using unpriviliged user to run Neo4j

jri-sp:
Just a proposal for running neo4j with a non-root user.
Even in a container, I think running applications with root user is not secure.

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -4,6 +4,13 @@ RUN apt-get update --quiet --quiet \
     && apt-get install --quiet --quiet --no-install-recommends lsof \
     && rm -rf /var/lib/apt/lists/*
 
+RUN gpg --keyserver ha.pool.sks-keyservers.net --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4
+RUN curl --output /usr/local/bin/gosu --fail --silent --show-error --location "https://github.com/tianon/gosu/releases/download/1.7/gosu-$(dpkg --print-architecture)" \
+        && curl --output /usr/local/bin/gosu.asc --fail --silent --show-error --location "https://github.com/tianon/gosu/releases/download/1.7/gosu-$(dpkg --print-architecture).asc" \
+        && gpg --verify /usr/local/bin/gosu.asc \
+        && rm /usr/local/bin/gosu.asc \
+        && chmod +x /usr/local/bin/gosu
+
 ENV NEO4J_VERSION %%VERSION%%
 ENV NEO4J_EDITION %%EDITION%%
 ENV NEO4J_DOWNLOAD_SHA256 %%DOWNLOAD_SHA%%
@@ -18,6 +25,9 @@ RUN curl --fail --silent --show-error --location --output neo4j.tar.gz $NEO4J_UR
     && tar --extract --file neo4j.tar.gz --directory /var/lib \
     && mv /var/lib/neo4j-* /var/lib/neo4j \
     && rm neo4j.tar.gz
+
+RUN useradd --system neo4j
+RUN chown --recursive neo4j: /var/lib/neo4j
 
 WORKDIR /var/lib/neo4j
 
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -10,6 +10,7 @@ setting() {
 }
 
 if [ "$1" == "neo4j" ]; then
+    chown --recursive neo4j: /data
     setting "keep_logical_logs" "${NEO4J_KEEP_LOGICAL_LOGS:-100M size}" neo4j.properties
     setting "dbms.pagecache.memory" "${NEO4J_CACHE_MEMORY:-512M}" neo4j.properties
     setting "wrapper.java.additional=-Dneo4j.ext.udc.source" "${NEO4J_UDC_SOURCE:-docker}" neo4j-wrapper.conf
@@ -22,7 +23,7 @@ if [ "$1" == "neo4j" ]; then
         setting "dbms.security.auth_enabled" "false" neo4j-server.properties
     elif [[ "${NEO4J_AUTH:-}" == neo4j/* ]]; then
         password="${NEO4J_AUTH#neo4j/}"
-        bin/neo4j start || \
+        exec gosu neo4j:neo4j bin/neo4j start || \
             (cat data/log/console.log && echo "Neo4j failed to start" && exit 1)
         if ! curl --fail --silent --user "neo4j:${password}" http://localhost:7474/db/data/ >/dev/null ; then
             curl --fail --silent --show-error --user neo4j:neo4j \
@@ -30,7 +31,7 @@ if [ "$1" == "neo4j" ]; then
                 --header 'Content-Type: application/json' \
                 http://localhost:7474/user/neo4j/password
         fi
-        bin/neo4j stop
+        exec gosu neo4j:neo4j bin/neo4j stop
     elif [ -n "${NEO4J_AUTH:-}" ]; then
         echo "Invalid value for NEO4J_AUTH: '${NEO4J_AUTH}'"
         exit 1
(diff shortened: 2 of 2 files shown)


benbc:
Hello @jri-sp. Thanks for the proposal and the code.

We are unhappy with running as root as well and plan to change it at some point. However I'm afraid it's not as simple as your commit makes it look.

The trouble comes when the `/data` volume is mounted on a directory in the host system that has restrictive permissions (as it is likely to do). In that case we need to `chown` the volume at container run time (i.e. in `docker-entrypoint.sh`) rather than build time (i.e. in `Dockerfile`).

Ordinarily we would use `sudo` for this, but the minimal Debian base that we use doesn't include `sudo` -- for good reason because it pulls in a lot of stuff and increases the attack surface significantly.

The standard approved Docker approach for this is exemplified by the solution in the Redis image ([here](https://github.com/docker-library/redis/tree/master/3.0) which uses a minimal `sudo`-analogue written by one of the Docker maintainers. (I'm not wild about this approach, but it seems to be the only way.)

In order to solve this properly we also need to set a value for `ulimit -n` which Neo4j needs to be a high value (>40k). Currently this is okay because we are running as `root`, but I suspect that it will need something explicit if we have a `neo4j` user.

If you want to tackle this then I'm very happy to give you feedback on a more complete pull request. Otherwise, I'm afraid that we don't have a schedule for doing it ourselves. It's unlikely to be in Q1 2016.

-Ben

jri-sp:
Hello @benbc ,

Thanks for the quick and consisting answer!

I'm glad that we share the same idea about using root user :) 
I know... my solution is minimalist, my pull request was more to open a discussion than accept it.

But maybe I can find time to work more on it, because I need it, so if I can help...

Thanks for the redis example, I didn't know this way to do, I'll study it!

For the /data folder, yes in my  configuration I'm using one and it's not working with my thing. I have to manualy chown my folder on the host side with uid/gid of the container, which is not a very user-friendly and beautiful solution...

However, for the ulimit, I've run my Docker image with --ulimit=nofile=40000:40000 and there is no warning in the logs. If I remove this parameter I have:
```
WARNING: Max 1024 open files allowed, minimum of 40 000 recommended. See the Neo4j manual.
```
So I was thinking it works,  but I'll investigate. 

You are more experienced than me in Docker, so thanks for all those advices!

Regards,

JRI

benbc:
@jri-sp Yes, I'm sure that `--ulimit=nofile` works correctly. But we'd rather not make the image harder to use by requiring that people add that parameter. So it needs something in the Dockerfile (or possibly `entrypoint-docker.sh`) that sets the value appropriately.

jri-sp:
Hi @benbc ,

I you have a few time to watch my thing...

I have done a new commit taking example on `redis` and `mongodb` Docker images using `gosu`.
Is it better for you ? 
There something you see that I didn't think ?

I don't touch the "NEO4J_AUTH" part because it's called temporary and I was thinking it's not a big deal if it's runned as root. No ?

So this version change the data volume's permissions mounted with -v option on the host system, which is more simple to use.

Best Regards,
jri

benbc:
Hello @jri-sp. Thanks so much for doing this work.

I've put a few comments inline, but they are mostly stylistic and basically I think the solution is good.

Can you please remove the first, obsolete commit from the PR?

I think we _do_ need to run the temporary Neo4j instance for the auth stuff as `neo4j`, for completeness and to help avoid problems with `root` owning files that `neo4j` needs to touch. That will mean moving the `chown` for `/data` up to the beginning of the block.

benbc:
I do think we need to address the `ulimit nofile` thing.

jri-sp:
Hello @benbc ,

Sorry for the closing/opening messages, actually I'm not familiar with Github and Pull Requests, so I probably made some mistakes and I've cleaned my first obsolete commit.

Thanks a lot for taking time to answer me!

I've applied all your comments and moved `chown` for /data to the top of Docker entry script to use `gosu` in the NEO4J_AUTH part.

Now, the interesting commit is c7a4496f8767c53a0f00e5542a46dd755163058c

For the ulimit, I need to do some tests to suggest something. I just know that is no possible to set that in the Dockerfile.

Thanks again for your help!

Best Regards,

JRI

jri-sp:
I've just done some tests for the ulimit. I don't have good news... :(

1) We can set it in Dockerfile
2) If we set it in docker-entry.sh, we got an "Operation not permitted"  error
```ulimit: open files: cannot modify limit: Operation not permitted```
It seems to be a docker security "feature": https://groups.google.com/forum/#!searchin/docker-user/limits/docker-user/T45Kc9vD804/v8J_N4gLbacJ

There is two solution:
```First method: remove sys_resource from lxc_template.go and recompile docker. Then you'll be able to ulimit as high as you like.
Second method: stop the docker demon. Use ulimit to increase the limit. Start the docker demon. It now has increased limits, and its child processes as well.```

By the way, I think setting it in the `docker run` command is the best solution, because:
- it's work.
- You can't ask to user to recompile their docker.
- Modifying limits on the host (second method), will apply the limit to all containers, which is not necessarily wanted and can cause issues.

I will continue my researches about this but I'm afraid to find nothing great...

**Comment by benbc** [comment]
@jri-sp This all looks great. Thanks for your patience with my picky comments. We need to do some research of our own about the `ulimit nofiles` thing before we can merge it.

**Comment by jri-sp** [comment]
@benbc  You're welcome. Let me know what you'll find about `ulimit`, as I said I'm beginning in the Docker world, maybe I've missed a info :)

**Comment by jri-sp** [comment]
Hi @benbc ,

I don't know if this PR is still relevant with Docker 1.10 and user namespaces... 
http://www.jrslv.com/docker-1-10/

Regards

**Comment by benbc** [comment]
Thanks for pointing that out. And thanks for your patience. I'm afraid this has got stuck behind other work at the moment.

**Comment by mneedham** [comment]
@jri-sp not sure if you're still following this thread but I think Docker engine has changed the default ulimit value (https://github.com/moby/moby/issues/25209). If I run the Neo4j image I see this:

```
$ docker run -it --rm --ulimit=nofile=1000:1000 neo4j sh -c "ulimit -n"
1000

$ docker run -it --rm neo4j sh -c "ulimit -n"
1048576
```

Do you still have your image handy to check whether you see the same behaviour even when running the container as non-root?

**Comment by jri-sp** [comment]
Hello @mneedham ,

I've rebuilt my image and it's OK too:
```
➜  ~ docker run -it --rm --ulimit=nofile=1000:1000 neo4j sh -c "ulimit -n"
1000
➜  ~ docker run -it --rm neo4j sh -c "ulimit -n"
1048576
➜  ~ docker run -it --rm jrisp/docker-neo4j sh -c "ulimit -n"
1048576
```

Tested on Docker for Mac:
```
➜  ~ docker version
Client:
 Version:      17.09.1-ce
 API version:  1.32
 Go version:   go1.8.3
 Git commit:   19e2cf6
 Built:        Thu Dec  7 22:22:25 2017
 OS/Arch:      darwin/amd64

Server:
 Version:      17.09.1-ce
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   19e2cf6
 Built:        Thu Dec  7 22:28:28 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

And CoreOS:
```
Client:
 Version:      17.09.0-ce
 API version:  1.32
 Go version:   go1.8.4
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:24:58 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.09.0-ce
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.4
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:24:58 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

If it can be useful, I've pulled my image with unprivileged user here: https://hub.docker.com/r/jrisp/docker-neo4j/

As I say before, I think this PR can be closed and forget, no ?

Moreover with last docker version, there is a way to map users so it's quite easy to change neo4j user:
```
➜  ~ docker run -it --rm --publish 7474:7474 --publish 7687:7687 --user 112 --name=neo4j  neo4j
➜  ~ docker exec -ti  neo4j ps aux
PID   USER     TIME   COMMAND
    1 112        0:19 /usr/lib/jvm/java-1.8-openjdk/jre/bin/java -cp /var/lib/n
  316 112        0:00 ps aux
```

Some related doc: https://docs.docker.com/engine/security/userns-remap/

I personally use this method (more tuned but this is what is done at the end) with my Neo4j in production with Docker.

But I'm curious to know your thoughts about this issue since I've submitted this PR :)

Regards

spacecowboy:
Closing as this has now been resolved by https://github.com/neo4j/docker-neo4j/pull/126

PR #23, opened 2016-01-20
Expose hook for downstream consumers of image

simpsonjulian:
If you're creating an image based on this one, you might want to
override 1 or 2 configuration values without duplicating the entire
entrypoint script.

This commit enables image builders to call an additional script via an
environment variable (if they add the script to their image).


simpsonjulian:
@benbc: I added some docs.

PR #27, opened 2016-02-17
Make start script more flexible to accept location

macmacbr:
Add a variable to describe an alternative to the root location for most
of the extra files:
    /conf
    /ssl
    /plubins
    _extension_script

all the above are prepended with an environment variable
ALTERNATE_DIR_LOCATION. The default value is empty. Wich does not change
previous behavior.

The ALTERNATE_DIR_LOCATION can point to a different location to find
those assets mentioned above, e.g., /my/extracted/zip/dir.

The variable can also be set to either another variable name, e.g.,
$MESOS_SANDBOX, in which case it will be evaluated to the content of the
later variable.

Also on this PR, the ability to set different ports for the cluster and
the server communication. Also quite useful for deployments on Mesos.

Latest, there's an improvement to the _settings_ function in order to be
able to append inexistant configurations.

### DIFF
--- a/2.3.2-enterprise/docker-entrypoint.sh
+++ b/2.3.2-enterprise/docker-entrypoint.sh
@@ -6,10 +6,18 @@ setting() {
     file="${3}"
     if [ -n "${value}" ]; then
         sed --in-place "s|.*${setting}=.*|${setting}=${value}|" conf/"${file}"
+        if ! grep "${setting}=" conf/"${file}" >/dev/null; then
+            echo "${setting}=${value}" >> conf/${file}
+        fi
     fi
 }
 
 if [ "$1" == "neo4j" ]; then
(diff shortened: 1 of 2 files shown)


benbc:
Hello @macmacbr. Thank you for this PR. It would be helpful if you can give some more background on why these changes are useful to you.

What does your use case look like without `ALTERNATE_DIR_LOCATION`?

What does the port configuration give you that you can't get by using Docker's port mapping?

macmacbr:
Hi benbc,

We are running neo4j in Marathon/Mesos. Marathon puts all its donwloaded files on a directory that changes depending on the slave and the instance. This PR is to support the ability to have any artifact downloaded being used by this script.

I was experimenting with the ports. Indeed we don't need to overwrite them. I can update the PR for that.

PR #35, opened 2016-06-02
Allow Setting Conf Option online_backup_server

alexanderkiel:
This commit enables the setting of online_backup_server using the env
var NEO4J_ONLINE_BACKUP_SERVER.


benbc:
@alexanderkiel It's a good idea to make it easy to expose online backup from the image; thank you. But I think we can make it even simpler.

Online backup is enabled by default in Neo4j Enterprise, but by default the backup server only binds to the loopback interface; to expose it remotely the operator needs to change the config to bind a different interface. I think the equivalent functionality for Docker is to unconditionally bind the server to `0.0.0.0:6362` and leave it up to the operator to specify a mapping for that port if they want to expose the functionality.

Do you agree?

I'd happily accept a PR that adds
```bash
    setting "online_backup_server" "0.0.0.0:6362" neo4j.properties
```
near the other network settings.

Please could you make the change to `docker-entrypoint.sh` in the root of the repo? The per-release subdirectories are added as part of our build process and aren't changed after the fact. There is a 2.3.x series release coming up before long which we may be able to get this change into.

benbc:
@alexanderkiel Are you still interested in this change?

alexanderkiel:
@benbc Yes and I have even a ToDo entry which I always push forward. Give me two weeks and lets see if I can manage it.

spacecowboy:
this was fixed by https://github.com/neo4j/docker-neo4j/pull/83

PR #40, opened 2016-06-20
Make lookup python3 compatible

spacecowboy:


benbc:
@spacecowboy This looks good to me.

PR #44, opened 2016-06-27
Remove all mention of obsolete config files

spacecowboy:
neo4j-wrapper.conf is still valid for 3.0. In 3.1, the settings function
correctly places settings in neo4j.conf instead (this is in fact
supported in 3.0 as well but let's keep to the official way).


benbc:
@spacecowboy This looks good to me. You have exercised commendable self-control in retaining the `neo4j-wrapper.conf` stuff for 3.0.

PR #45, opened 2016-06-27
Separate arg for initial heap size

MkhIr:


benbc:
@MkhIr Can you tell me why this would be useful to you? Our experience is that it's easier to manage Neo4j if the entire heap is allocated immediately.

benbc:
@MkhIr You only need to change `docker-entrypoint.sh` in the root of the codebase as we won't retrospectively change already-released images.

MkhIr:
@benbc  in an experiment I was making It was causing issue when assigning more that 75% of the sys ram (in my case 52 of 64). Having separate arg would allow allocation of acceptable number with a max heap for cases such as loading large csv or match.

benbc:
@MkhIr It sounds like your envisaging a case where you'll run an operation (e.g. loading large CSV) that may or may not work at random depending on what else is going on on the machine at the time. It's avoiding exactly this kind of uncertainty that leads us to recommend setting a fixed size for the heap.

benbc:
@MkhIr Are you still interested in this change?

MkhIr:
no you can ignore it if thats the case.

PR #46, opened 2016-06-29
Set settings that aren't already present

pvanheus:
The current version of setting() only edits settings already present in the config files. I discovered this while trying to set a read_only setting but it holds true in general. The version included sets appends settings if they are not already present in the config file. If this change is accepted it would have to be applied to all copies of the entrypoint script.

Peter

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -5,7 +5,11 @@ setting() {
     value="${2}"
     file="${3}"
     if [ -n "${value}" ]; then
-        sed --in-place "s|.*${setting}=.*|${setting}=${value}|" conf/"${file}"
+        if  grep -q ".*${setting}=" conf/"${file}" ; then
+            echo "${setting}=${value}" >> conf/"${file}"
+        else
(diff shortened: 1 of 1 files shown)


benbc:
@pvanheus Thank you for this PR. You're right that we should support this case.

Perhaps I am misreading the code, but isn't the logic back-to-front in the implementation. You want to do the `sed` in the case that the `grep` succeeds and the `echo` otherwise.

A couple of minor quibbles on the code:

* please use the full form of `grep`'s `-q` flag for readability
* please redirect `grep`'s stdout to `/dev/null` so that nothing gets printed in the case where the setting is already present

benbc:
@pvanheus Are you still interested in this change?

benbc:
@pvanheus It sounds like you don't need this any more. I'm going to close this for now. Please reopen the PR and have a look at my comments above if are still interested in getting the change in.

benbc:
@pvanheus I ran into the need to do this myself, so I have implemented it in c8dbffb3e8bf9e568aa12138bde26574167062b0. Hope that you find it useful.

PR #53, opened 2016-09-28
Listen on localhost only when setting credentials

DeltaWhy:
When `NEO4J_AUTH` is set, the server is started to set the credentials, then stopped and started again with the final configuration. When a container is restarted, this initial server listens on `0.0.0.0` causing other dependent containers to see the Neo4j container as ready before it actually is. This PR fixes the issue by explicitly setting the address to `127.0.0.1` for the server pre-run.

Previous discussion at https://github.com/neo4j/neo4j/issues/7203


spacecowboy:
Thanks for the PR @DeltaWhy 
sorry about the delay

DeltaWhy:
Thanks! When can I expect to see this change pushed to Docker Hub?

benbc:
In the next milestone. This week or next.

PR #60, opened 2016-10-14
Update heap defaults to include size unit

spacecowboy:
This [PR](https://github.com/neo4j/neo4j/pull/8138) enables the use of
units in heap config.

`neo4j-wrapper.conf` is deprecated so no reason to use it.


spacecowboy:
@benbc changed that and simplified the setting function

benbc:
Looks great. Thanks.

PR #63, opened 2016-11-01
Use shorthand flags to that work on OSX.

maxsumrall:


benbc:
Thanks, Max.

benbc:
@maxsumrall I've taken off the changelog label. We shouldn't be changelogging stuff that is just build infrastructure -- only changes to the actual product surface.

PR #73, opened 2016-12-20
Example fix for grep failures causing early exit

matthew-cox:
Hey Ben!

I've encountered an error when building an image derived from the published Neo4j image. 

Specifically, my attempt to replace config files seems to interfere with the files which are expected to exist. This, in turn, causes grep to fail (file not found) when the setting function is called and aborts the execution of docker-entrypoint.sh.

I've changed one of the docker-entrypoint.sh scripts. If you think this is a reasonable approach (one possible improvement might be put an explicit check for file existence) I will be happy to change the other scripts as well.

Thank you!
Matt

### DIFF
--- a/src/2.3/docker-entrypoint.sh
+++ b/src/2.3/docker-entrypoint.sh
@@ -6,11 +6,17 @@ setting() {
     file="${3}"
 
     if [ -n "${value}" ]; then
+        # I think the intent is to always set *something*, disable exit on error which allows:
+        # 1) Success if the file doesn't exist
+        # 2) Success if grep exits with non-0 for any other reason
+        set +e
         if grep --quiet --fixed-strings "${setting}=" conf/"${file}"; then
             sed --in-place "s|.*${setting}=.*|${setting}=${value}|" conf/"${file}"
         else
             echo "${setting}=${value}" >>conf/"${file}"
         fi
+        # restore it
+        set -e
     fi
 }
 


benbc:
Hello Matt. I have either misunderstood the problem you're seeing or the fix you are proposing.

The failure of the `grep` invocation should never cause an exit, irrespective of whether `-e` is set, because the `if` swallows the error and triggers the `else` branch.

There is a minor problem here, though, which may have put you onto the wrong path: when the file is missing there will be a message printed to stderr. I would definitely like to fix that by redirecting `grep`'s stderr to `/dev/null`.

matthew-cox:
Morning Ben!

Of course it's now been a couple weeks and I can no longer reproduce what I was seeing repeatedly. 

It's possible I had something else going wrong. Let me poke at it a bit more and get back to you.

Matt

matthew-cox:
Ok, I found it.

My use of `set -e` does fix things, but it's not the `grep` which is failing, it's the `echo` command which fails:

    $ grep -v '^#' Dockerfile
    FROM neo4j:2.3
    WORKDIR /conf
    COPY neo4j* ./
    
    $ docker build .
    Sending build context to Docker daemon 24.58 kB
    Step 1 : FROM neo4j:2.3
     ---> 01913b7bb5ba
    Step 2 : WORKDIR /conf
     ---> Using cache
     ---> 3c7bb41d5678
    Step 3 : COPY neo4j* ./
     ---> Using cache
     ---> cdd5c75b4555
    Step 4 : CMD neo4j
     ---> Running in c93a6e636a38
     ---> 853e5d07964b
    Removing intermediate container c93a6e636a38
    Successfully built 853e5d07964b
    
    $ docker run 853e5d07964b
    grep: conf/neo4j.properties: No such file or directory
    /docker-entrypoint.sh: line 12: conf/neo4j.properties: No such file or directory
    
    $ docker run --entrypoint=/bin/bash -ti 853e5d07964b
    
    root@03f82e8e5094:/# head -12 /docker-entrypoint.sh | tail -1
                echo "${setting}=${value}" >>conf/"${file}"

However, it's the use of a relative path which causes the failure. If I alter my Dockerfile and add `WORKDIR /var/lib/neo4j` at the end: it works.

    $ grep -v '^#' Dockerfile
    FROM neo4j:2.3
    WORKDIR /conf
    COPY neo4j* ./
    WORKDIR /var/lib/neo4j
    
    $ docker build --no-cache .
    Sending build context to Docker daemon 24.58 kB
    Step 1 : FROM neo4j:2.3
     ---> 01913b7bb5ba
    Step 2 : WORKDIR /conf
     ---> Running in 8ec4e2482eb8
     ---> 920792a54af7
    Removing intermediate container 8ec4e2482eb8
    Step 3 : COPY neo4j* ./
     ---> 7215d9fd81ab
    Removing intermediate container 8efb4e859441
    Step 4 : WORKDIR /var/lib/neo4j
     ---> Running in 5b64f36e3de6
     ---> d0d0029a2c82
    Removing intermediate container 5b64f36e3de6
    Successfully built d0d0029a2c82
    
    $ docker run d0d0029a2c82
    Starting Neo4j Server console-mode...
    2017-01-04 18:13:31.040+0000 INFO  No SSL certificate found, generating a self-signed certificate..
    2017-01-04 18:13:35.124+0000 INFO  Successfully started database
    2017-01-04 18:13:35.152+0000 INFO  Starting HTTP on port 7474 (16 threads available)
    2017-01-04 18:13:35.355+0000 INFO  Enabling HTTPS on port 7473
    2017-01-04 18:13:35.402+0000 WARN  Acceptors should be <= availableProcessors: ServerConnector@670672a9{HTTP/1.1}{0.0.0.0:0}
    2017-01-04 18:13:35.427+0000 WARN  Acceptors should be <= availableProcessors: ServerConnector@3ef59e0e{SSL-HTTP/1.1}{0.0.0.0:0}
    2017-01-04 18:13:35.460+0000 INFO  Mounting static content at /webadmin
    2017-01-04 18:13:35.528+0000 INFO  Mounting static content at /browser
    2017-01-04 18:13:37.054+0000 INFO  Remote interface ready and available at http://0.0.0.0:7474/

I'm not sure this is a bug per se, but it seems a bit odd that the entry script lives at `/` but needs to be executed from within `/var/lib/neo4j` to work.

I can work around by altering my `Dockerfile`. I'm open to discussing other solutions too.

Thank you,
Matt

PR #80, opened 2017-03-19
add setting "dbms.directories.import" "/import" neo4j.conf

stdob:
Right now we need to specify the directory for import directly. For example:

        -v $HOME/neo4j/import:/var/lib/neo4j/import \

instead:

        -v $HOME/neo4j/import:/import \


spacecowboy:
Thanks for the PR. This looks reasonable and consistent with the other directories.

PR #83, opened 2017-04-03
Configuration from not hardcoded env variables

### LINKED ISSUE #52: dbms.backup.address config value cannot be overridden by an env variable
I would like to initialize Neo4J backups remotely but obviously can't without enabling the 'dbms.backup.address' config value.  I have had to take a copy of the neo4j.conf and inject it to the docker container because the value cannot be overridden using an environment variable like other settings can.

Could this be fixed?  Ideally, so that all settings within the config file could be overridden.

jangaraj:
Problem: only hardcoded env variables can be used to configure neo4j container. It's only subset of all possible settings

Solution: introduce specific env variable naming convention, which will be used for dynamic configuration

Some things to keep in mind:
- Some settings are commented by default and some are not
- Not all settings are actually present (commented or not) in the config file. So a sed by itself is not guaranteed to replace something
- A user might define a custom config volume AND some environment variables.
- Backward compatibility with old hardcoded env variables
- Update documentation/examples

WIP: it's only for 3.2 version at the moment; it needs review from somebody with proper neo4j config knowledge (cc: @spacecowboy); I was not able to build docker image on my localhost

This one can be a solution for: #52, #71.

Fixes #52 
Fixes #71

### DIFF
--- a/src/3.2/Dockerfile
+++ b/src/3.2/Dockerfile
@@ -4,8 +4,8 @@ RUN apk add --no-cache --quiet \
     bash \
     curl
 
-ENV NEO4J_SHA256 %%NEO4J_SHA%%
-ENV NEO4J_TARBALL %%NEO4J_TARBALL%%
+ENV NEO4J_SHA256 %%NEO4J_SHA%% \
+    NEO4J_TARBALL %%NEO4J_TARBALL%%
 ARG NEO4J_URI=%%NEO4J_DIST_SITE%%/%%NEO4J_TARBALL%%
 
 COPY ./local-package/* /tmp/
@@ -14,12 +14,11 @@ RUN curl --fail --silent --show-error --location --remote-name ${NEO4J_URI} \
     && echo "${NEO4J_SHA256}  ${NEO4J_TARBALL}" | sha256sum -csw - \
     && tar --extract --file ${NEO4J_TARBALL} --directory /var/lib \
     && mv /var/lib/neo4j-* /var/lib/neo4j \
-    && rm ${NEO4J_TARBALL}
+    && rm ${NEO4J_TARBALL} \
+    && mv /var/lib/neo4j/data /data \
+    && ln -s /data /var/lib/neo4j/data
 
 WORKDIR /var/lib/neo4j
-
-RUN mv data /data \
-    && ln -s /data
 
 VOLUME /data
 
--- a/src/3.2/Dockerfile
+++ b/src/3.2/Dockerfile
@@ -4,8 +4,8 @@ RUN apk add --no-cache --quiet \
     bash \
     curl
 
-ENV NEO4J_SHA256 %%NEO4J_SHA%% \
-    NEO4J_TARBALL %%NEO4J_TARBALL%%
+ENV NEO4J_SHA256=%%NEO4J_SHA%% \
+    NEO4J_TARBALL=%%NEO4J_TARBALL%%
 ARG NEO4J_URI=%%NEO4J_DIST_SITE%%/%%NEO4J_TARBALL%%
 
 COPY ./local-package/* /tmp/
@@ -16,7 +16,8 @@ RUN curl --fail --silent --show-error --location --remote-name ${NEO4J_URI} \
     && mv /var/lib/neo4j-* /var/lib/neo4j \
     && rm ${NEO4J_TARBALL} \
     && mv /var/lib/neo4j/data /data \
-    && ln -s /data /var/lib/neo4j/data
+    && ln -s /data /var/lib/neo4j/data \
+    && apk del curl
 
 WORKDIR /var/lib/neo4j
 
--- a/src/3.2/docker-entrypoint.sh
+++ b/src/3.2/docker-entrypoint.sh
@@ -80,6 +80,10 @@ if [ "$1" == "neo4j" ]; then
         NEO4J_dbms_directories_import="/import"
     fi
 
+    if [ -d /metrics ]; then
+        NEO4J_dbms_directories_metrics="/metrics"
+    fi
+
     if [ "${NEO4J_AUTH:-}" == "none" ]; then
         NEO4J_dbms_security_auth__enabled=false
     elif [[ "${NEO4J_AUTH:-}" == neo4j/* ]]; then
--- a/src/3.2/docker-entrypoint.sh
+++ b/src/3.2/docker-entrypoint.sh
@@ -44,21 +44,21 @@ if [ "$1" == "neo4j" ]; then
         NEO4J_causalClustering_raftAdvertisedAddress
 
     # Custom settings for dockerized neo4j
-    NEO4J_dbms_tx__log_rotation_retention_policy=${NEO4J_dbms_tx__log_rotation_retention_policy:-100M size}
-    NEO4J_dbms_memory_pagecache_size=${NEO4J_dbms_memory_pagecache_size:-512M}
-    NEO4J_wrapper_java_additional=${NEO4J_wrapper_java_additional:--Dneo4j.ext.udc.source=docker}
-    NEO4J_dbms_memory_heap_initial__size=${NEO4J_dbms_memory_heap_initial__size:-512M}
-    NEO4J_dbms_memory_heap_max__size=${NEO4J_dbms_memory_heap_max__size:-512M}
-    NEO4J_dbms_connectors_default__listen__address=${NEO4J_dbms_connectors_default__listen__address:-0.0.0.0}
-    NEO4J_dbms_connector_http_listen__address=${NEO4J_dbms_connector_http_listen__address:-0.0.0.0:7474}
-    NEO4J_dbms_connector_https_listen__address=${NEO4J_dbms_connector_https_listen__address:-0.0.0.0:7473}
-    NEO4J_dbms_connector_bolt_listen__address=${NEO4J_dbms_connector_bolt_listen__address:-0.0.0.0:7687}
-    NEO4J_causal__clustering_discovery__listen__address=${NEO4J_causal__clustering_discovery__listen__address:-0.0.0.0:5000}
-    NEO4J_causal__clustering_discovery__advertised__address=${NEO4J_causal__clustering_discovery__advertised__address:-$(hostname):5000}
-    NEO4J_causal__clustering_transaction__listen__address=${NEO4J_causal__clustering_transaction__listen__address:-0.0.0.0:6000}
-    NEO4J_causal__clustering_transaction__advertised__address=${NEO4J_causal__clustering_transaction__advertised__address:-$(hostname):6000}
-    NEO4J_causal__clustering_raft__listen__address=${NEO4J_causal__clustering_raft__listen__address:-0.0.0.0:7000}
-    NEO4J_causal__clustering_raft__advertised__address=${NEO4J_causal__clustering_raft__advertised__address:-$(hostname):7000}
+    : ${NEO4J_dbms_tx__log_rotation_retention_policy:-100M size}
+    : ${NEO4J_dbms_memory_pagecache_size:-512M}
+    : ${NEO4J_wrapper_java_additional:--Dneo4j.ext.udc.source=docker}
+    : ${NEO4J_dbms_memory_heap_initial__size:-512M}
+    : ${NEO4J_dbms_memory_heap_max__size:-512M}
+    : ${NEO4J_dbms_connectors_default__listen__address:-0.0.0.0}
+    : ${NEO4J_dbms_connector_http_listen__address:-0.0.0.0:7474}
+    : ${NEO4J_dbms_connector_https_listen__address:-0.0.0.0:7473}
+    : ${NEO4J_dbms_connector_bolt_listen__address:-0.0.0.0:7687}
+    : ${NEO4J_causal__clustering_discovery__listen__address:-0.0.0.0:5000}
+    : ${NEO4J_causal__clustering_discovery__advertised__address:-$(hostname):5000}
+    : ${NEO4J_causal__clustering_transaction__listen__address:-0.0.0.0:6000}
+    : ${NEO4J_causal__clustering_transaction__advertised__address:-$(hostname):6000}
+    : ${NEO4J_causal__clustering_raft__listen__address:-0.0.0.0:7000}
+    : ${NEO4J_causal__clustering_raft__advertised__address:-$(hostname):7000}
 
     if [ -d /conf ]; then
         find /conf -type f -exec cp {} conf \;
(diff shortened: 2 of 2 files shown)


src/3.2/Dockerfile
```diff
@@ -4,8 +4,8 @@ RUN apk add --no-cache --quiet \
     bash \
     curl
 
-ENV NEO4J_SHA256 %%NEO4J_SHA%%
-ENV NEO4J_TARBALL %%NEO4J_TARBALL%%
+ENV NEO4J_SHA256 %%NEO4J_SHA%% \
+    NEO4J_TARBALL %%NEO4J_TARBALL%%
```
spacecowboy: This change makes the image not build

src/3.2/docker-entrypoint.sh
```diff
@@ -1,30 +1,77 @@
...
+    NEO4J_dbms_memory_pagecache_size=${NEO4J_dbms_memory_pagecache_size:-512M}
+    NEO4J_wrapper_java_additional=${NEO4J_wrapper_java_additional:--Dneo4j.ext.udc.source=docker}
+    NEO4J_dbms_memory_heap_initial__size=${NEO4J_dbms_memory_heap_initial__size:-512M}
+    NEO4J_dbms_memory_heap_max__size=${NEO4J_dbms_memory_heap_max__size:-512M}
+    NEO4J_dbms_connectors_default__listen__address=${NEO4J_dbms_connectors_default__listen__address:-0.0.0.0}
+    NEO4J_dbms_connector_http_listen__address=${NEO4J_dbms_connector_http_listen__address:-0.0.0.0:7474}
+    NEO4J_dbms_connector_https_listen__address=${NEO4J_dbms_connector_https_listen__address:-0.0.0.0:7473}
+    NEO4J_dbms_connector_bolt_listen__address=${NEO4J_dbms_connector_bolt_listen__address:-0.0.0.0:7687}
+    NEO4J_causal__clustering_discovery__listen__address=${NEO4J_causal__clustering_discovery__listen__address:-0.0.0.0:5000}
+    NEO4J_causal__clustering_discovery__advertised__address=${NEO4J_causal__clustering_discovery__advertised__address:-$(hostname):5000}"
```
spacecowboy: Inconsistent use of quotation marks here. The rest of the file becomes quoted and fails a syntax check

src/3.2/docker-entrypoint.sh
```diff
@@ -38,47 +85,19 @@ if [ "$1" == "neo4j" ]; then
...
-    setting "causal_clustering.discovery_advertised_address" "${NEO4J_causalClustering_discoveryAdvertisedAddress:-$(hostname):5000}"
-    setting "causal_clustering.transaction_listen_address" "${NEO4J_causalClustering_transactionListenAddress:-0.0.0.0:6000}"
-    setting "causal_clustering.transaction_advertised_address" "${NEO4J_causalClustering_transactionAdvertisedAddress:-$(hostname):6000}"
-    setting "causal_clustering.raft_listen_address" "${NEO4J_causalClustering_raftListenAddress:-0.0.0.0:7000}"
-    setting "causal_clustering.raft_advertised_address" "${NEO4J_causalClustering_raftAdvertisedAddress:-$(hostname):7000}"
+    # list env variables with prefix NEO4J_ and create settings from them
+    for i in $( printenv | grep ^NEO4J_ | awk -F'=' '{print $1}' | sort -rn ); do
+        setting=$(echo ${i} | sed 's|^NEO4J_||' | sed 's|_|.|g' | sed 's|\.\.|_|g')
+        value=$(echo ${!i})
+        if grep -q -F "^${setting}=" conf/"${file}"; then
```
spacecowboy: `conf/${file}` gives an error for unbound variable. I think you want to hardcode this to `conf/neo4j.conf`

src/3.2/docker-entrypoint.sh
```diff
@@ -38,47 +85,19 @@ if [ "$1" == "neo4j" ]; then
...
+            sed --in-place "s|.*${setting}=.*|${setting}=${value}|" conf/neo4j.conf
+        else
+            echo "${setting}=${value}" >> conf/"${file}"
```
spacecowboy: Same, change to `conf/neo4j.conf`

src/3.2/docker-entrypoint.sh
spacecowboy: Could this grep use the same regex as the sed below? E.g. make it also find commented values?

Another option than replacing in-place would be to remove (or comment) the existing entry and simply fall through to the other case and always append the environment value to the end of the file.

I'm fine with either.

spacecowboy:
To build and run the tests do this:

```
echo "NEO4J_VERSION=3.2.0-alpha07" > devenv.local && source devenv && make test-community
```

If you start a new terminal, remember to source `devenv` again.

You will see in the console output what the image id is, e.g.

```
Successfully built 9c2fa1709590
```

So you can do some manual tests as well. This is needed since the tests don't really tell you why they fail (bash syntax errors for instance)

src/3.2/docker-entrypoint.sh
jangaraj: Fixed.

src/3.2/docker-entrypoint.sh
jangaraj: Done. Hardcoded "conf/neo4j.conf"

jangaraj: I agree. I've used the same regex as the sed below.

src/3.2/docker-entrypoint.sh
jangaraj: Done. Hardcoded "conf/neo4j.conf"

src/3.2/Dockerfile
spacecowboy: Reference: https://docs.docker.com/engine/reference/builder/#env

Use equal signs instead of space to fix this problem:

```
ENV NEO4J_SHA256=%%NEO4J_SHA%% \
    NEO4J_TARBALL=%%NEO4J_TARBALL%%
```

spacecowboy:
@jangaraj I made some changes and the tests all ran successfully with them. 

* Changed the `ENV` directive to use equal signs to support several variables
* Changed declaration of default values to a shorter form
* Loop over all variables using `set` instead of `printenv` or locally defined `NEO4J_` vars are not included
* Do not set anything if `$value` is null.

Here's the diff:

```diff
diff --git i/src/3.2/Dockerfile w/src/3.2/Dockerfile
index f158142..619ee8b 100644
--- i/src/3.2/Dockerfile
+++ w/src/3.2/Dockerfile
@@ -4,8 +4,8 @@ RUN apk add --no-cache --quiet \
     bash \
     curl
 
-ENV NEO4J_SHA256 %%NEO4J_SHA%% \
-    NEO4J_TARBALL %%NEO4J_TARBALL%%
+ENV NEO4J_SHA256=%%NEO4J_SHA%% \
+    NEO4J_TARBALL=%%NEO4J_TARBALL%%
 ARG NEO4J_URI=%%NEO4J_DIST_SITE%%/%%NEO4J_TARBALL%%
 
 COPY ./local-package/* /tmp/
diff --git i/src/3.2/docker-entrypoint.sh w/src/3.2/docker-entrypoint.sh
index 3f879f3..615a47c 100755
--- i/src/3.2/docker-entrypoint.sh
+++ w/src/3.2/docker-entrypoint.sh
@@ -44,21 +44,21 @@ if [ "$1" == "neo4j" ]; then
         NEO4J_causalClustering_raftAdvertisedAddress
 
     # Custom settings for dockerized neo4j
-    NEO4J_dbms_tx__log_rotation_retention_policy=${NEO4J_dbms_tx__log_rotation_retention_policy:-100M size}
-    NEO4J_dbms_memory_pagecache_size=${NEO4J_dbms_memory_pagecache_size:-512M}
-    NEO4J_wrapper_java_additional=${NEO4J_wrapper_java_additional:--Dneo4j.ext.udc.source=docker}
-    NEO4J_dbms_memory_heap_initial__size=${NEO4J_dbms_memory_heap_initial__size:-512M}
-    NEO4J_dbms_memory_heap_max__size=${NEO4J_dbms_memory_heap_max__size:-512M}
-    NEO4J_dbms_connectors_default__listen__address=${NEO4J_dbms_connectors_default__listen__address:-0.0.0.0}
-    NEO4J_dbms_connector_http_listen__address=${NEO4J_dbms_connector_http_listen__address:-0.0.0.0:7474}
-    NEO4J_dbms_connector_https_listen__address=${NEO4J_dbms_connector_https_listen__address:-0.0.0.0:7473}
-    NEO4J_dbms_connector_bolt_listen__address=${NEO4J_dbms_connector_bolt_listen__address:-0.0.0.0:7687}
-    NEO4J_causal__clustering_discovery__listen__address=${NEO4J_causal__clustering_discovery__listen__address:-0.0.0.0:5000}
-    NEO4J_causal__clustering_discovery__advertised__address=${NEO4J_causal__clustering_discovery__advertised__address:-$(hostname):5000}
-    NEO4J_causal__clustering_transaction__listen__address=${NEO4J_causal__clustering_transaction__listen__address:-0.0.0.0:6000}
-    NEO4J_causal__clustering_transaction__advertised__address=${NEO4J_causal__clustering_transaction__advertised__address:-$(hostname):6000}
-    NEO4J_causal__clustering_raft__listen__address=${NEO4J_causal__clustering_raft__listen__address:-0.0.0.0:7000}
-    NEO4J_causal__clustering_raft__advertised__address=${NEO4J_causal__clustering_raft__advertised__address:-$(hostname):7000}
+    : ${NEO4J_dbms_tx__log_rotation_retention_policy:=100M size}
+    : ${NEO4J_dbms_memory_pagecache_size:=512M}
+    : ${NEO4J_wrapper_java_additional:=-Dneo4j.ext.udc.source=docker}
+    : ${NEO4J_dbms_memory_heap_initial__size:=512M}
+    : ${NEO4J_dbms_memory_heap_max__size:=512M}
+    : ${NEO4J_dbms_connectors_default__listen__address:=0.0.0.0}
+    : ${NEO4J_dbms_connector_http_listen__address:=0.0.0.0:7474}
+    : ${NEO4J_dbms_connector_https_listen__address:=0.0.0.0:7473}
+    : ${NEO4J_dbms_connector_bolt_listen__address:=0.0.0.0:7687}
+    : ${NEO4J_causal__clustering_discovery__listen__address:=0.0.0.0:5000}
+    : ${NEO4J_causal__clustering_discovery__advertised__address:=$(hostname):5000}
+    : ${NEO4J_causal__clustering_transaction__listen__address:=0.0.0.0:6000}
+    : ${NEO4J_causal__clustering_transaction__advertised__address:=$(hostname):6000}
+    : ${NEO4J_causal__clustering_raft__listen__address:=0.0.0.0:7000}
+    : ${NEO4J_causal__clustering_raft__advertised__address:=$(hostname):7000}
 
     if [ -d /conf ]; then
         find /conf -type f -exec cp {} conf \;
@@ -97,13 +97,15 @@ if [ "$1" == "neo4j" ]; then
 
     # list env variables with prefix NEO4J_ and create settings from them
     unset NEO4J_AUTH NEO4J_SHA256 NEO4J_TARBALL
-    for i in $( printenv | grep ^NEO4J_ | awk -F'=' '{print $1}' | sort -rn ); do
-        setting=$(echo ${i} | sed 's|^NEO4J_||' | sed 's|_|.|g' | sed 's|\.\.|_|g')
-        value=$(echo ${!i})
-        if grep -q -F "${setting}=" conf/neo4j.conf; then
+    for i in $( set | grep ^NEO4J_ | awk -F'=' '{print $1}' | sort -rn ); do
+      setting=$(echo ${i} | sed 's|^NEO4J_||' | sed 's|_|.|g' | sed 's|\.\.|_|g')
+      value=$(echo ${!i})
+      if [[ -n ${value} ]]; then
+          if grep -q -F "${setting}=" conf/neo4j.conf; then
             sed --in-place "s|.*${setting}=.*|${setting}=${value}|" conf/neo4j.conf
-        else
+          else
             echo "${setting}=${value}" >> conf/neo4j.conf
+          fi
         fi
     done

```

jangaraj:
@spacecowboy nice bash skills 👍 
- `printenv` vs `set` - I'm not sure about set, I have had some problems with special characters in password environment variables in my projects - it needs some testing with `NEO4J_AUTH`
- I've removed also curl - IMHO it's required only for docker image build process
- there are some settings `dbms.directories.*` (for example `lib`), which are not included in the entrypoint script  volumes part - do we want to implement them as well?

spacecowboy:
@jangaraj 

> there are some settings dbms.directories.* (for example lib), which are not included in the entrypoint script volumes part - do we want to implement them as well?

Definitely not the `lib` dir. It should only point the proper installation folder. It only make sense to override it if you install neo4j according to some odd directory structure. And the user shouldn't concern themselves where those things live inside the container.

> I've removed also curl - IMHO it's required only for docker image build process

That's probably perfectly fine

>  I'm not sure about set, I have had some problems with special characters in password environment variables in my projects - it needs some testing with NEO4J_AUTH

Wasn't aware of any difference. Good catch. Might be worth adding a test for this. I'll do it as soon as you push your latest changes.

jangaraj:
@spacecowboy 
Yes, volume support for `dbms.directories.lib` setting doesn't looks like a good idea. I've added /metrics volume support.

I'm not sure about meaning of this line in the Dockerfile:
```
COPY ./local-package/* /tmp/
```
Do we really need it?

I'm still not able to build docker image locally:
```
[root@dockerhost docker-neo4j]# echo "NEO4J_VERSION=3.2.0-alpha07" > devenv.local && source devenv &&  make test-community
bash: - : invalid option 
bash: - : invalid option
bash: - : invalid option 
make: *** No rule to make target `tmp/.image-id-community', needed by `tmp/.tests-pass-community'.  Stop.
```
Could you publish test docker image with this change please?

src/3.2/docker-entrypoint.sh
```diff
@@ -1,30 +1,91 @@
...
+    : ${NEO4J_dbms_connectors_default__listen__address:-0.0.0.0}
+    : ${NEO4J_dbms_connector_http_listen__address:-0.0.0.0:7474}
+    : ${NEO4J_dbms_connector_https_listen__address:-0.0.0.0:7473}
+    : ${NEO4J_dbms_connector_bolt_listen__address:-0.0.0.0:7687}
+    : ${NEO4J_causal__clustering_discovery__listen__address:-0.0.0.0:5000}
+    : ${NEO4J_causal__clustering_discovery__advertised__address:-$(hostname):5000}
+    : ${NEO4J_causal__clustering_transaction__listen__address:-0.0.0.0:6000}
+    : ${NEO4J_causal__clustering_transaction__advertised__address:-$(hostname):6000}
+    : ${NEO4J_causal__clustering_raft__listen__address:-0.0.0.0:7000}
+    : ${NEO4J_causal__clustering_raft__advertised__address:-$(hostname):7000}
```
spacecowboy: All of these need to use `:=` instead of `:-`

For example:
```
: ${NEO4J_dbms_memory_heap_max__size:=512M}
```

src/3.2/docker-entrypoint.sh
```diff
@@ -1,30 +1,91 @@
...
+
+    if [ -d /logs ]; then
+        NEO4J_dbms_directories_logs="/logs"
+    fi
+
+    if [ -d /import ]; then
+        NEO4J_dbms_directories_import="/import"
+    fi
+
+    if [ -d /metrics ]; then
```
spacecowboy: :+1:

spacecowboy:
Please keep this for now:

```
COPY ./local-package/* /tmp/
```

I think it's used somewhere in the build pipeline.

jangaraj:
Done. BTW: `spacecowboy/neo4j:pr83` is bigger than `neo4j:latest` (249MB vs 172MB). That step `COPY ./local-package/* /tmp/` has copied `73.5M /tmp/neo4j-community-3.2.0-alpha07-unix.tar.gz` into image.

If you need additional modification, please create fork of my fork https://github.com/jangaraj/docker-neo4j/tree/env-variables (unfortunately I'm more less offline this week, so it'll be faster solution).

spacecowboy:
@jangaraj the size thing is something we can look at in a separate PR. This looks good here.

Pushed an updated build to `spacecowboy/neo4j:pr83` and all the tests pass. Will merge this soon (we are in the middle of a release, waiting til after to avoid issues)

PR #89, opened 2017-05-17
Ensure config only has single entries

spacecowboy:
This fixes a bug where dbms.mode would be set multiple times in the
config. Turns out that this is fine for all values EXCEPT
dbms.mode=ARBITER.

In that case we rely on parsing the config from bash which runs into a
problem due to:
https://github.com/neo4j/neo4j/blob/6cc2ebb93b39405976801f57173079deea45fd11/packaging/standalone/src/main/distribution/shell-scripts/bin/neo4j-shared.m4#L157

If an entry is defined multiple times, it is regarded as a cumulative
option. This is used for JAVA_OPTS and similarly. But for Arbiter we
set a different main class. So having multiple dmbs.mode entries will
set an incorrect value which means the default main class is called,
instead of the Arbiter's.

So this fix makes sure to only replace uncommented values in the
config, and otherwise always append to the end. Previously we would
replace even commented values, but since dmbs.mode is listed twice
in the 3.1/3.2/3.3 config (once in the HA-section, once in CC-section)
it would get set twice.

changelog: Fixed issue which prevented `dbms.mode=ARBITER` from working

### DIFF
--- a/src/3.0/docker-entrypoint.sh
+++ b/src/3.0/docker-entrypoint.sh
@@ -133,10 +133,11 @@ if [ "$1" == "neo4j" ]; then
         value=$(echo ${!i})
         if [[ -n ${value} ]]; then
             if grep -q -F "${setting}=" conf/neo4j.conf; then
-                sed --in-place "s|.*${setting}=.*|${setting}=${value}|" conf/neo4j.conf
-            else
-               echo "${setting}=${value}" >> conf/neo4j.conf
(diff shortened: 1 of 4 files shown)


spacecowboy:
Made sure to remove existing lines from the config because HEAP_SIZE was parsed twice on restart, would likely be the case for others as well

spacecowboy:
There is no 3.3 version to test with yet so ignore that failure

PR #96, opened 2017-07-04
Listening address for the bolt connector

stdob:
Set the value of the listening address for the bolt connector through the env variable 

> NEO4J_dbms_connector_bolt_listenAddress


nicorikken:
Why hasn't anyone looked at this yet? I need this feature too, but I'll probably resort to a custom config for now. I can't even see the output of the builds, so I can't fix the failures. Please Neo4j team, open up to community collaboration.

jennyowen:
@nicorikken it's been possible to set bolt listening port via environment variable for a while now. 
For example to change it to `7777`:
```
docker run -it --rm \
--publish=7474:7474 --publish=7777:7777 \
-e NEO4J_dbms_connector_bolt_listen__address=:7777 \
neo4j:3.5
```
Which gives this output:
```
Active database: graph.db
Directories in use:
  home:         /var/lib/neo4j
  config:       /var/lib/neo4j/conf
  logs:         /logs
  plugins:      /var/lib/neo4j/plugins
  import:       /var/lib/neo4j/import
  data:         /var/lib/neo4j/data
  certificates: /var/lib/neo4j/certificates
  run:          /var/lib/neo4j/run
Starting Neo4j.
2019-11-27 12:49:02.960+0000 INFO  ======== Neo4j 3.5.12 ========
2019-11-27 12:49:02.970+0000 INFO  Starting...
2019-11-27 12:49:04.818+0000 INFO  Bolt enabled on 0.0.0.0:7777
2019-11-27 12:49:05.822+0000 INFO  Started.
2019-11-27 12:49:06.607+0000 INFO  Remote interface available at http://localhost:7474/
```

Overriding configuration settings is documented here:
https://neo4j.com/docs/operations-manual/current/docker/configuration/

nicorikken:
@jennyowen Thanks for the quick response!

That is not the option I was looking for. I'm fine with port 7687, as long as I can change the advertised port in the GUI. As I use Kubernetes to switch around the ports to my liking. But I need bolt to be hosted on 443 due to cluster-wide policies (I actually use 2 different hostst: 1 for the GUI, 1 for bolt, both serving 443).

I could use the suggested listening address to get the UI to publish to 443 directly. But I'd have to change some port numbers around in my Kubernetes config as internal services expect the bolt to be available on 7687 still.

For now I've worked around it using this 'hack', only modifying the advertised address in the GUI:
```
        command:
          - "/bin/bash"
          - "-c"
          - |
            echo "dbms.connector.bolt.advertised_address={{ .Values.ingress.bolt.host}}:443" >> conf/neo4j.conf
            exec /docker-entrypoint.sh "neo4j"
```

Either way there might be cases where the container port for bolt might be different than the port by which the UI has to connect. This PR adresses that, and just changing the listening address doesn't.

Is this a use-case you are willing to support? Otherwise this PR might as well be closed after more than 2 years of inactivity.

Issue #94, opened 2017-06-13
Allow for passing sensitive settings through files


gkapkowski:
As in some images provided on Docker Hub allow for setting `*_FILE` environment variables for passing sensitive data like passwords.

Example from PostgreSQL https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L8

spacecowboy:
This is already possible in a way. 

You can specify any neo4j configuration option via environment variables (https://neo4j.com/docs/operations-manual/current/installation/docker/#docker-environment-variables) and `docker run` supports `--env-file` argument to set environment variables from a file.
 (https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e-env-env-file)

The security implications seem slightly better here than in the approach taken by postgres:

* Host environment: You have a file with the password in it (both approaches)
* Docker environment: You have an environment variable with the password in it (both approaches)
* Docker environment: You have a file with the password in it (only postgres approach)

but I might be missing something else. I'm not sure what 

> [...] Docker's secrets feature)

is referring to and how that operates.

gkapkowski:
Yeah, I was thinking about --env-file but my current setup doesn't allow me to do it easily.

I'm using Rancher and their Secrets (https://docs.rancher.com/rancher/v1.6/en/cattle/secrets/) implementation allows only mounted /run/secrets/mysecret (inside container) files at the moment. 

Kubernetes (https://kubernetes.io/docs/concepts/configuration/secret/) has both options for file + environment variable so it will be working even now.

Ideal solution for me would be for images to allow passing sensitive data through files mounted by docker secrets. This would lead to situation when the only place sensitive data is visible is inside docker container through file (where file permissions could be used to further secure it).

Feel free to close this issue. I'll write a wrapper that will suite my needs for now. Hopefully some standard way of dealing with secrets will emerge soon.

PR #107, opened 2017-10-30
#94 Allow auth credentials to be configured via files.

josephtaylor:
This allows for the auth credentials to be specified via a file so docker secrets can be used.

So you could define a neo4j service like:
```yaml
version: '3.1'

secrets:
  neo4j-auth:
    external: true

services:
  neo4j:
    command: /docker-entrypoint.sh neo4j
    deploy:
      mode: replicated
      replicas: 1
    environment:
      NEO4J_AUTH_FILE: /run/secrets/neo4j-auth
```

Where `neo4j-auth` is a file that has the contents: `neo4j/somefancypassword`

------------------

The `file_env` function was lifted from the mysql entrypoint script:
https://github.com/docker-library/mysql/blob/master/5.7/docker-entrypoint.sh#L21-L41

### DIFF
--- a/src/2.3/docker-entrypoint.sh
+++ b/src/2.3/docker-entrypoint.sh
@@ -14,6 +14,28 @@ setting() {
     fi
 }
 
+# usage: file_env VAR [DEFAULT]
+#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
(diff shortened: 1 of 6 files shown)


kcrawley:
it's worth noting that mariadb, along with postgres implement the exact same behavior.

PR #123, opened 2018-03-12
Dev jgrobbel docker on mac

jgrobbel:
This PR allows make to complete successfully on a mac.
The key fixes are:
1. Reducing the memory footprint of the containers.
2. Use /tmp for mktemp files as /var is not available to docker
3. Use a jump host network container to give access to the docker0 bridge.

### DIFF
--- a/devenv
+++ b/devenv
@@ -2,6 +2,51 @@
 
 PATH="./build:${PATH}"
 
+# MacOS specific checks
+if [ "$(uname)" == "Darwin" ] ; then
+
+  echo "Setting PATH with MacOS specific locations"
+  export PATH="/usr/local/opt/coreutils/libexec/gnubin:$PATH"
+  export PATH="/usr/local/opt/gnu-sed/libexec/gnubin:$PATH"
+  export PATH="/usr/local/opt/make/libexec/gnubin:$PATH"
+
+  if ! grep --version 2>/dev/null | grep -q "GNU grep" ; then
+    cat >&2 <<EOF
+== GNU grep is required. Fix with ==
+
+  brew install grep --with-default-names
+
+EOF
+  fi
+
+  if ! sed --version 2>/dev/null | grep -q "GNU sed" ; then
+    cat >&2 <<EOF
+== GNU sed is required. Fix with ==
+
+  brew install gnu-sed
+
+EOF
+  fi
+
+  if ! readlink --version 2>/dev/null | grep -q "GNU coreutils" ; then
+    cat >&2 <<EOF
+== GNU coreutils is required. Fix with ==
+
+  brew install coreutils  --with-default-names
+
+EOF
+  fi
+
+  # have to use || here because BSD find doesnt even support --version (it errors)
+  # shellcheck disable=SC2185
+  find --version 1>/dev/null 2>/dev/null || cat >&2 <<EOF
+== GNU findutils is required. Fix with ==
+
+    brew install findutils  --with-default-names
+
+EOF
+fi # MacOS specific checks
+
 if ! which make >/dev/null; then
     if which apt-get >/dev/null; then
         sudo apt-get install make
--- a/test/causal-cluster-compose.yml
+++ b/test/causal-cluster-compose.yml
@@ -1,4 +1,4 @@
-version: '2'
+version: '3'
 
 networks:
   lan:
@@ -12,6 +12,8 @@ services:
       - lan
     environment:
       - NEO4J_ACCEPT_LICENSE_AGREEMENT=yes
+      - NEO4J_dbms_memory_pagecache_size=10M
+      - NEO4J_dbms_memory_heap_initial__size=10M
       - NEO4J_AUTH=none
       - NEO4J_dbms_mode=CORE
       - NEO4J_causalClustering_expectedCoreClusterSize=3
@@ -23,6 +25,8 @@ services:
       - lan
     environment:
       - NEO4J_ACCEPT_LICENSE_AGREEMENT=yes
+      - NEO4J_dbms_memory_pagecache_size=10M
+      - NEO4J_dbms_memory_heap_initial__size=10M
       - NEO4J_AUTH=neo4j/neo
       - NEO4J_dbms_mode=CORE
       - NEO4J_causalClustering_expectedCoreClusterSize=3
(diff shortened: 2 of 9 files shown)


devenv
```diff
@@ -2,6 +2,60 @@
...
+
+    brew install findutils  --with-default-names
+
+EOF
+
+  if grep "osxkeychain" ~/.docker/config.json ; then
+    cat >&2 <<EOF
+== Docker configured with keychain ==
+
+  Remove osxkeychain from ~/.docker/config.json
```
eastlondoner: I don't think you need this unless you are planning on having docker interact with private repositories in automation and we don't do that in this repository AFAIK

test/causal-cluster-compose.yml
```diff
@@ -12,8 +12,11 @@ services:
       - lan
     environment:
       - NEO4J_ACCEPT_LICENSE_AGREEMENT=yes
-      - NEO4J_AUTH=none
+      - NEO4J_AUTH=neo4j/neo
       - NEO4J_dbms_mode=CORE
+      - NEO4J_causalClustering_discoveryAdvertisedAddress=core1:5000
+      - NEO4J_causalClustering_transactionAdvertisedAddress=core1:6000
+      - NEO4J_causalClustering_raftAdvertisedAddress=core1:7000
```
eastlondoner: IMO this test is explicitly checking that the default behaviour hasn't regressed. I don't think these extra AdvertisedAddress values should be added.

test/container/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine
+
+RUN apk update && apk add --no-cache curl bash util-linux grep
+
+CMD ["bash", "-c", "while true; do sleep 120; done"]
```
eastlondoner: `sleep infinity` is also valid

test/causal-cluster-compose.yml
```diff
@@ -12,8 +12,11 @@ services:
       - lan
     environment:
       - NEO4J_ACCEPT_LICENSE_AGREEMENT=yes
-      - NEO4J_AUTH=none
+      - NEO4J_AUTH=neo4j/neo
```
eastlondoner: I think this causes the test-sets-password test to fail

test/causal-cluster-compose.yml
spacecowboy: @eastlondoner is correct. these settings were purposefully left out of the first core to make sure that default settings are good. They are included in subsequent cores to verify that they can be overwritten.

test/causal-cluster-compose.yml
spacecowboy: Don't change this value

test/test-dumps-config
```diff
@@ -7,7 +7,8 @@ readonly image="$1"
 readonly series="$2"
 readonly cname="neo4j-$(uuidgen)"
 
-readonly dir=$(mktemp --directory)
+# Docker for mac doesn't allow access from /var/
```
spacecowboy: this comment is strange. you mean on Mac `mktemp` creates temporary files in `/var`?

devenv
jgrobbel: ack

test/causal-cluster-compose.yml
jgrobbel: ack

test/causal-cluster-compose.yml
jgrobbel: ack

test/test-dumps-config
jgrobbel: Yep:
```
mktemp --directory
/var/folders/wc/zvf2ggjj1x12lvbpqjl9cs3m0000gn/T/tmp.8Kge0Cklu0
```

src/2.3/docker-entrypoint.sh
```diff
@@ -16,9 +16,9 @@ setting() {
 
 if [ "$1" == "neo4j" ]; then
     setting "keep_logical_logs" "${NEO4J_KEEP_LOGICAL_LOGS:-100M size}" neo4j.properties
-    setting "dbms.pagecache.memory" "${NEO4J_CACHE_MEMORY:-512M}" neo4j.properties
```
spacecowboy: Running with `10M` page cache is fine during cluster tests but don't change the production default

src/3.3/docker-entrypoint.sh
```diff
@@ -94,9 +94,9 @@ unset NEO4J_dbms_txLog_rotation_retentionPolicy NEO4J_UDC_SOURCE \
 
 # Custom settings for dockerized neo4j
 : ${NEO4J_dbms_tx__log_rotation_retention__policy:=100M size}
-: ${NEO4J_dbms_memory_pagecache_size:=512M}
+: ${NEO4J_dbms_memory_pagecache_size:=10M}
```
spacecowboy: Don't change production default

src/3.4/docker-entrypoint.sh
```diff
@@ -87,9 +87,9 @@ unset NEO4J_dbms_txLog_rotation_retentionPolicy NEO4J_UDC_SOURCE \
 
 # Custom settings for dockerized neo4j
 : ${NEO4J_dbms_tx__log_rotation_retention__policy:=100M size}
-: ${NEO4J_dbms_memory_pagecache_size:=512M}
+: ${NEO4J_dbms_memory_pagecache_size:=10M}
```
spacecowboy: Don't change production default

test/test-dumps-config
spacecowboy: That's insane. How do you put up with these toy machines?

test/causal-cluster-compose.yml
jgrobbel: I'll drop this commit

test/container/Dockerfile
jgrobbel: Sadly only supported by gnu sleep (so not on alpine or mac even)

jgrobbel:
@eastlondoner @spacecowboy 
I think I have addressed all the comments now.

test/test-dumps-config
benbc: @jgrobbel This comment and the one below are not sufficiently complete to be helpful. The pertinent in formation is that on OSX `mktemp` creates files under `/var` _and_ that that is not accessible from within Docker containers.

jgrobbel: Ack that.

You can configure it to work (which is more painful than just using the base config) but then there is a risk of mounting the hosting VM into the containers which will end badly.
https://github.com/docker/for-mac/issues/1532

test/ha-cluster-compose.yml
```diff
@@ -12,6 +12,8 @@ services:
       - lan
     environment:
       - NEO4J_ACCEPT_LICENSE_AGREEMENT=yes
+      - NEO4J_dbms_memory_pagecache_size=10M
```
spacecowboy: :+1:

PR #125, opened 2018-03-13
Run Neo4j as non-root 

### LINKED ISSUE #50: neo4j shouldn't run as root
if neo4j has any exploits this would allow the container to be exploited, including updating neo4js binary, or host other malign software within the container. This also makes it easier to attempt to break out of the jail which has been accomplished in the past.

spacecowboy:
This builds on #22

It uses `su-exec` [instead of gosu](https://github.com/tianon/gosu#su-exec) as
recommended by the author of `gosu` himself for Alpine images.

I have confirmed that docker these days sets a sensible ulimit so this
is no longer an issue we need to consider.

There is a fair amount of special handling in the entrypoint for
file permissions. We chown the relevant directories to the neo4j user
to make them writable for the process but we take care not to chown
directories which are volumes (to reasonable extents).

If a volume is mounted in `/data` or `/conf` (in that order) the neo4j
user is created with the same UID and GID as those directories. This
has the effect that the container will write new files with the same
UID/GID as the user owning the host directory. So once the container
exists, files will not have surprising (and maybe non-existing) owners on the host system.

Tests have been added to verify that the UID of files written are
sensible when volume mounts exist.

Fixes #50

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ tmp/.tests-pass-%: tmp/.image-id-% $(shell find test -name 'test-*') $(shell fin
 > mkdir -p $(@D)
 > image_id=$$(cat $<)
 > for test in $(filter test/test-%,$^); do
->   echo "Running $${test}"
+>   echo "Running $${test} with ${series} $*"
 >   "$${test}" "$${image_id}" "${series}" "$*"
 > done
 > touch $@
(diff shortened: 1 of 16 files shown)


spacecowboy:
Closing and reopening because of a CI issue

PR #126, opened 2018-03-14
Run Neo4j as non-root 

spacecowboy:
This builds on #22

It uses `su-exec` [instead of gosu](https://github.com/tianon/gosu#su-exec) as
recommended by the author of `gosu` himself for Alpine images.

I have confirmed that docker these days sets a sensible ulimit so this
is no longer an issue we need to consider.

There is a fair amount of special handling in the entrypoint for
file permissions. We chown the relevant directories to the neo4j user
to make them writable for the process but we take care not to chown
directories which are volumes (to reasonable extents).

If a volume is mounted in `/data` or `/conf` (in that order) the neo4j
user is created with the same UID and GID as those directories. This
has the effect that the container will write new files with the same
UID/GID as the user owning the host directory. So once the container
exists, files will not have surprising (and maybe non-existing) owners on the host system.

Tests have been added to verify that the UID of files written are
sensible when volume mounts exist.

### DIFF
--- a/src/3.1/docker-entrypoint.sh
+++ b/src/3.1/docker-entrypoint.sh
@@ -63,16 +63,14 @@ done <   <(find /var/lib/neo4j -type d -mindepth 1 -maxdepth 1 -print0)
 
 # Data dir is chowned later
 
-if [[ "${cmd}" != *"neo4j"* ]]; then
-  if [ "${cmd}" == "dump-config" ]; then
-    if [ -d /conf ]; then
-      # Run with neo4j user so we write files with correct permissions
-      su-exec "${user_name}" cp --recursive conf/* /conf
-      exit 0
-    else
-      echo >&2 "You must provide a /conf volume"
-      exit 1
-    fi
+if [ "${cmd}" == "dump-config" ]; then
+  if [ -d /conf ]; then
+    # Run with neo4j user so we write files with correct permissions
+    su-exec "${user_name}" cp --recursive conf/* /conf
+    exit 0
+  else
+    echo >&2 "You must provide a /conf volume"
+    exit 1
   fi
 fi
 
--- a/src/3.1/docker-entrypoint.sh
+++ b/src/3.1/docker-entrypoint.sh
@@ -3,19 +3,15 @@
 cmd="$1"
 
 if [[ "${cmd}" != *"neo4j"* ]]; then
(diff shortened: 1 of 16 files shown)


test/test-starts-up-with-data-volume
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset
+
+. "$(dirname "$0")/helpers.sh"
+
+readonly image="$1"
+readonly series="$2"
+readonly cname="neo4j-$(uuidgen)"
+
+readonly datadir=$(mktemp --directory)
```
eastlondoner: please add `--tmpdir=/tmp/` for OS X compatibility 🍏

spacecowboy: Ah right. Done

src/3.1/docker-entrypoint.sh
```diff
@@ -3,19 +3,15 @@
 cmd="$1"
 
 if [[ "${cmd}" != *"neo4j"* ]]; then
```
benbc: This outer test can go altogether, can't it? (And in the other branches.)

src/3.1/docker-entrypoint.sh
```diff
@@ -2,10 +2,76 @@
...
+if [[ "$(stat -c %u /data)" != "0" ]]; then
+  # /data is a volume, is the same uid/gid for neo4j user
+  user_uid="$(stat -c %u /data)"
+  user_gid="$(stat -c %g /data)"
+elif [ -d /conf ] && [[ "${cmd}" == "dump-config" ]]; then
+  # A configuration volume has been mounted and we are dumping config
+  user_uid="$(stat -c %u /conf)"
+  user_gid="$(stat -c %g /conf)"
+else
+  # Set to zero (root) to signal no mounting has been done
```
benbc: Using `0` as a sentinel value here is misleading, since we never actually want to use `root`. Can you just leave them unset and test against that instead?

spacecowboy: Actually, it serves a dual-purpose. In the special case that a user would have mounted a volume which was actually a directory owned by root, then this approach will make sure we do in fact run as a non-root user. So using `0` as the sentinel value makes sure we never run under that uid.

We can of course add an additional check to the ifs later to check if it's `unset or zero`. If you think that's cleaner I'll change it.

Writing the reply has convinced me that it's cleaner so I'll change it...

PR #129, opened 2018-03-20
Simplify non root execution

spacecowboy:
Neo4j will always run as non-root. But if the user wants sensible UIDs
on the files in any volume, the user needs to make sure to invoke
docker with the `--user` option.

This matches what other official images do to run as non-root.

### DIFF
--- a/src/3.1/Dockerfile
+++ b/src/3.1/Dockerfile
@@ -1,10 +1,6 @@
 FROM openjdk:8-jre-alpine
 
-RUN apk add --no-cache --quiet \
-    bash \
-    curl \
-    tini \
-    su-exec
+RUN addgroup -S neo4j && adduser -S -H -h /var/lib/neo4j -G neo4j neo4j
 
 ENV NEO4J_SHA256=%%NEO4J_SHA%% \
     NEO4J_TARBALL=%%NEO4J_TARBALL%%
@@ -12,15 +8,23 @@ ARG NEO4J_URI=%%NEO4J_DIST_SITE%%/%%NEO4J_TARBALL%%
 
 COPY ./local-package/* /tmp/
 
-RUN curl --fail --silent --show-error --location --remote-name ${NEO4J_URI} \
+RUN apk add --no-cache --quiet \
+    bash \
+    curl \
+    tini \
+    su-exec \
+    && curl --fail --silent --show-error --location --remote-name ${NEO4J_URI} \
     && echo "${NEO4J_SHA256}  ${NEO4J_TARBALL}" | sha256sum -csw - \
     && tar --extract --file ${NEO4J_TARBALL} --directory /var/lib \
     && mv /var/lib/neo4j-* /var/lib/neo4j \
     && rm ${NEO4J_TARBALL} \
     && mv /var/lib/neo4j/data /data \
-    && chown -R root:root /data \
-    && chown -R root:root /var/lib/neo4j \
-    && ln -s /data /var/lib/neo4j/data
+    && chown -R neo4j:neo4j /data \
+    && chmod -R 777 /data \
+    && chown -R neo4j:neo4j /var/lib/neo4j \
+    && chmod -R 777 /var/lib/neo4j \
+    && ln -s /data /var/lib/neo4j/data \
+    && apk del curl
 
 ENV PATH /var/lib/neo4j/bin:$PATH
 
--- a/src/3.1/docker-entrypoint.sh
+++ b/src/3.1/docker-entrypoint.sh
@@ -2,62 +2,38 @@
 
 cmd="$1"
 
-# We intentionally chown /data to root in Dockerfile so we can detect
-# if no volume is mounted
-if [[ "$(stat -c %u /data)" != "0" ]]; then
-  # /data is a volume, is the same uid/gid for neo4j user
-  user_uid="$(stat -c %u /data)"
-  user_gid="$(stat -c %g /data)"
-elif [ -d /conf ] && [[ "${cmd}" == "dump-config" ]]; then
-  # A configuration volume has been mounted and we are dumping config
-  user_uid="$(stat -c %u /conf)"
-  user_gid="$(stat -c %g /conf)"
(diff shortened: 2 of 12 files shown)


src/3.1/Dockerfile
```diff
@@ -1,26 +1,31 @@
...
-    bash \
-    curl \
-    tini \
-    su-exec
+RUN addgroup -S neo4j && adduser -S -H -h /var/lib/neo4j -G neo4j neo4j
 
 ENV NEO4J_SHA256=%%NEO4J_SHA%% \
-    NEO4J_TARBALL=%%NEO4J_TARBALL%%
+    NEO4J_TARBALL=%%NEO4J_TARBALL%% \
+    NEO4J_EDITION=%%NEO4J_EDITION%%
```
benbc: Why is `NEO4J_EDITION` added here?

src/3.1/docker-entrypoint.sh
```diff
@@ -2,73 +2,42 @@
...
 # Data dir is chowned later
 
 if [ "${cmd}" == "dump-config" ]; then
   if [ -d /conf ]; then
-    # Run with neo4j user so we write files with correct permissions
-    # Note that su-exec, despite its name, does not replicate the
-    # functionality of exec, so we need to use both
-    exec su-exec "${user_name}" cp --recursive conf/* /conf
+    # Assume user is running docker with --user
+    exec cp --recursive conf/* /conf
```
benbc: I can see that this is safe, because if we're `root` then for a read-only operation like this becoming `neo4j` doesn't gain us anything. But the comment begs questions. Would it be more straight-forward to do the `su-exec` dance here as well?

src/3.1/docker-entrypoint.sh
```diff
@@ -184,17 +153,22 @@ done
...
+  exec su-exec neo4j neo4j console
+elif [ "${cmd}" == "neo4j" ]; then
+  # docker running with --user
+  exec neo4j console
+elif [ "$(id -u)" = "0" ]; then
+  exec su-exec neo4j "$@"
 else
-    exec su-exec "${user_name}" "$@"
+  # docker running with --user
+  exec "$@"
```
benbc: I suggest simplifying all this by creating an `${exec_cmd}` variable in the `if [ "$(id -u)" = "0" ]` switch right at the top and then applying that unconditionally to each of these cases. Would that work?

src/3.1/Dockerfile
spacecowboy: Oh, copy paste mistake

src/3.1/docker-entrypoint.sh
spacecowboy: if you prefer, sure

src/3.1/docker-entrypoint.sh
spacecowboy: can try

src/3.1/docker-entrypoint.sh
```diff
@@ -2,73 +2,46 @@
...
 fi
 
 while IFS= read -r -d '' dir
 do
-  if [[ "$(stat -c %u "${dir}")" = "0" ]]; then
+  if [[ "$(stat -c %U "${dir}")" = "neo4j" ]]; then
     # Using mindepth 1 to avoid the base directory here so recursive is OK
-    chown -R "${user_name}:${group_name}" "${dir}"
+    chown -R "${userid}":"${groupid}" "${dir}" || :
+    chmod -R 700 "${dir}" || :
```
benbc: What are the failure cases here?

src/3.1/docker-entrypoint.sh
```diff
@@ -184,17 +157,16 @@ done
 
 # Chown the data dir now that (maybe) an initial password has been
 # set (this is a file in the data dir)
-if [[ "$(stat -c %u /data)" = "0" ]]; then
-  chown -R "${user_name}:${group_name}" /data
-fi
+chown -R "${userid}":"${groupid}" /data || :
+chmod -R 700 /data || :
```
benbc: And here?

src/3.1/docker-entrypoint.sh
```diff
@@ -184,17 +157,16 @@ done
...
 
 # Use su-exec to drop privileges to neo4j user
 # Note that su-exec, despite its name, does not replicate the
 # functionality of exec, so we need to use both
 if [ "${cmd}" == "neo4j" ]; then
-    exec su-exec "${user_name}" neo4j console
+  ${exec_cmd} neo4j console
 else
-    exec su-exec "${user_name}" "$@"
+  ${exec_cmd} "$@"
```
benbc: Huge improvement. Thank you.

src/3.1/docker-entrypoint.sh
spacecowboy: User mounts a volume in `/var/lib/neo4j/conf` (we have a test for this) but does not specify a `--user` which matches the owner of the directory.

In that case, the directory might be writable even - but the user inside of docker does not have permission to chmod or chown it

src/3.1/docker-entrypoint.sh
spacecowboy: mounted data volume has an owner which does not match the user the docker container is running as. The directory needs to be writable for the executing user, but we don't have permission to reown or chmod it in that case.

src/3.1/docker-entrypoint.sh
benbc: And there is no reasonable way to test for that case specifically?

spacecowboy: Yes there is

```
if [[ "$(stat -c %u /path/to)" = "${userid}" ]]; then
  # we have permisssion to chown/chmod
fi
```

But line-wise it's longer

benbc: I think that's preferable. Otherwise someone reading the code is always left wondering how else it could fail.

And for the one below as well, if possible.

spacecowboy: OK, changed.

PR #137, opened 2018-05-30
Don't write things that look like neo4j config environment variables …

eastlondoner:
but start with a number to the neo4j.conf

It causes neo4j to fail with very confusing errors.

N.b. this would be a problem if in future we had conf settings that start with numbers but that seems unlikely since A) it would cause the neo4j shell script to barf and B) it doesn't fit our naming convention.

full list of config settings is here for reference
https://neo4j.com/docs/operations-manual/current/reference/configuration-settings/


benbc:
@eastlondoner I thought that we had a discussion somewhere about outputting a warning in case there is an env var being ignored by this -- otherwise someone is going to have a difficult time diagnosing a problem at some point.

PR #143, opened 2018-06-12
Allow tmp root to vary per system

benbc:
mktemp uses $TMPDIR as the tmp root if it is set and defaults to /tmp
if it is not. So there is no reason to specify --tmpdir=/tmp. On Linux
TMPDIR is not set by default, and /tmp will be used. On OSX it is set
and uses some user-specific location. On TeamCity $TMPDIR is set to a
build-specific location.

It's actually the TeamCity behaviour that we're doing this for. Using
/tmp there is problematic because that is very small on our build
agents, so we need to use $TMPDIR instead.

### DIFF
--- a/devenv
+++ b/devenv
@@ -5,6 +5,20 @@ PATH="./build:${PATH}"
 # MacOS specific checks
 if [ "$(uname)" == "Darwin" ] ; then
 
+  # We use temporary directories during testing that need to be
+  # accessible to the Docker daemon. By default on Macs, $TMPDIR
+  # (which mktemp responds to) is set to a user-specific location that
+  # the Docker daemon cannot read from.
+  #
+  # In some environments, such as TeamCity, $TMPDIR is intentionally
+  # pointed elsewhere, so we only want to override the default
+  # value. (We don't currently run these builds on Macs, but you never
+  # know.) This default seems to be in /private/var/folders on some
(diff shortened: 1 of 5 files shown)


benbc:
[TC builds for the docker image are using the /tmp directory. They shouldn't because it is not inside the build's enclosure. ./tmp is much better.](https://trello.com/c/7KaSHKxb/142-tc-builds-for-the-docker-image-are-using-the-tmp-directory-they-shouldnt-because-it-is-not-inside-the-builds-enclosure-tmp-is-mu)

eastlondoner:
we could consider setting TMPDIR on Mac to `/tmp` in devenv.
It seems Mac uses /var/folders/<random> for $TMPDIR

benbc:
@eastlondoner `/tmp` is a symlink on Macs, I  think. To `/private/tmp`? Will Docker be able to write there happily?

jgrobbel:
@benbc Yes /tmp should be fine. We use it in the docker repo:

```
ls -l /tmp
lrwxr-xr-x@ 1 root  wheel  11 21 Oct  2017 /tmp -> private/tmp
```

```
# mktemp on OSX by default uses /var/folders/ which is not available to docker
readonly dir=$(mktemp --directory --tmpdir=/tmp/)
```

Issue #152, opened 2018-10-03
instructions on website to run docker don't work with last version


jeromecc:
Hello!
Instructions are here: https://neo4j.com/developer/docker/

docker run \
    --publish=7474:7474 --publish=7687:7687 \
    --volume=$HOME/neo4j/data:/data \
    --volume=$HOME/neo4j/logs:/logs \
    neo4j:3.0

This works well.

But:
docker run \
    --publish=7474:7474 --publish=7687:7687 \
    --volume=$HOME/neo4j/data:/data \
    --volume=$HOME/neo4j/logs:/logs \
    neo4j
Doesn't. This is pulling the latest version which is 3.4.7 at time of writing.

The instructions in the README.md file works well:
docker run \
    --publish=7474:7474 --publish=7687:7687 \
    --volume=$HOME/neo4j/data:/data \
    neo4j:3.0

Someone should update the website.

Thank you for the amazing work!

Cheers,

jérôme

jennyowen:
Thanks for reporting this.
It looks as though the logging may have changed in 3.1. If I don't specify subfolders it works as expected:
```
docker run 
--publish=7474:7474 --publish=7687:7687 
--volume=$HOME/neo4j:/data 
--volume=$HOME/neo4j:/logs 
neo4j
```
I can update the README here, it looks like our docs also have the broken code in. 
/cc @davidoliverSP2

jennyowen:
On further inspection, I think the documentation is actually correct.
This looks to be caused by file permissions on the non-containerised mount point for the `/logs`. Mounting `/data` seems to work file. 
The `docker-entrypoint.sh` is setting file ownership to `neo4j:neo4j` on `/data` but it misses out doing that for `/logs`.

jennyowen:
Fixed with https://github.com/neo4j/docker-neo4j/pull/161
This will be in all future Neo4j  docker releases, so from these versions and onwards:
- 3.5.3
- 3.4.13
- 3.3.10

This isn't a guarantee that we'll make those specific releases. More that if we release those versions, you'll be able to mount the `logs` directory in the same way as our documentation.

PR #161, opened 2019-01-23
fixed mounting of /logs folder and added test for fail.

jennyowen:
Fix for issue https://github.com/neo4j/docker-neo4j/issues/152

### DIFF
--- a/src/3.1/Dockerfile
+++ b/src/3.1/Dockerfile
@@ -20,12 +20,12 @@ RUN apk add --no-cache --quiet \
     && rm ${NEO4J_TARBALL} \
     && mv /var/lib/neo4j/data /data \
     && chown -R neo4j:neo4j /data \
-    && chmod -R 755 /data \
+    && chmod -R 777 /data \
     && mv /var/lib/neo4j/logs /logs \
     && chown -R neo4j:neo4j /logs \
-    && chmod -R 755 /logs \
+    && chmod -R 777 /logs \
     && chown -R neo4j:neo4j /var/lib/neo4j \
-    && chmod -R 755 /var/lib/neo4j \
+    && chmod -R 777 /var/lib/neo4j \
     && ln -s /data /var/lib/neo4j/data \
     && ln -s /logs /var/lib/neo4j/logs \
     && apk del curl
--- a/src/3.1/Dockerfile
+++ b/src/3.1/Dockerfile
@@ -20,12 +20,12 @@ RUN apk add --no-cache --quiet \
     && rm ${NEO4J_TARBALL} \
     && mv /var/lib/neo4j/data /data \
     && chown -R neo4j:neo4j /data \
-    && chmod -R 777 /data \
+    && chmod -R 755 /data \
     && mv /var/lib/neo4j/logs /logs \
     && chown -R neo4j:neo4j /logs \
-    && chmod -R 777 /logs \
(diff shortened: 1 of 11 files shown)


src/3.1/Dockerfile
```diff
@@ -21,16 +21,21 @@ RUN apk add --no-cache --quiet \
     && mv /var/lib/neo4j/data /data \
     && chown -R neo4j:neo4j /data \
     && chmod -R 777 /data \
+    && mv /var/lib/neo4j/logs /logs \
+    && chown -R neo4j:neo4j /logs \
+    && chmod -R 777 /logs \
```
eastlondoner: why are we using `777` here but `755` later?

src/3.1/docker-entrypoint.sh
```diff
@@ -165,6 +165,9 @@ done
 if [[ "$(id -u)" = "0" ]]; then
   chmod -R 755 /data
   chown -R "${userid}":"${groupid}" /data
+  # Chown the log dir also
+  chmod -R 755 /logs
+  chown -R "${userid}":"${groupid}" /logs
```
eastlondoner: This change could break things for existing users who have a valid setup. if `/logs` is mounted from outside then this will change the owner on the directory on the host machine!

Are we definitely happy to do that? 

I think it would probably be better to check if the user can write to the logs directory and if they can't to exit with an error.

To be honest I think we should do the same with `/data` (i.e. fail fast with an error instead of changing the owner). But since that would break for existing users we can't in `3.x`.

test/test-starts-up-with-data-and-log-volumes
```diff
@@ -0,0 +1,62 @@
...
+  "${l_image}")"
+  echo "log: tmp/out/${cid}.log"
+  trap "docker_cleanup ${cid}" EXIT
+}
+
+readonly image="$1"
+readonly series="$2"
+readonly cname="neo4j-$(uuidgen)"
+
+readonly datadir=$(mktemp --directory)
```
eastlondoner: can you name this variable `data_and_logs_dir` or something that shows it contains both the data and the logs.

src/3.1/docker-entrypoint.sh
jennyowen: I don't think there's a way of creating a writable folder on the host that the `neo4j` user in the container has write permissions to. Other than passing the user/group ids into `docker run` anyway.

Even if I `chown` the folder on the host to `neo4j:neo4j` it doesn't work. Probably because the host user neo4j is different to the container user neo4j. :woman_shrugging: 

I'm not sure what the best solution is. We could chown it but print a warning that we did so?

test/test-starts-up-with-data-and-log-volumes
jennyowen: done!

src/3.1/Dockerfile
jennyowen: use 755 everywhere now

src/3.1/docker-entrypoint.sh
jennyowen: I don't think there's an easy solution to this. The solution with the `chown` works for our data directory and is how redis solve the problem of shared file access from a container: https://github.com/docker-library/redis/blob/master/5.0/Dockerfile

Checking we can write to a folder will always fail, unless container is run with a specific user and group.

Issue #164, opened 2019-03-18
entrypoint is making files and folders world writable and executable


rweverwijk:
Hi All,

Both Dockerfile and the docker-entrypoint.sh are changing the rights of a some folders recursively. This is giving some unwanted security risks.

There is no reason to permit the world write and execute access to `/data` or `/logs` and all it's files.

My suggestion would be to change the access in the docker file to:

- 700 for `/data` and sub directories (read not the files)
- 700 for `/data/databases`
- 755 for `/logs`
- 700 for `${NEO4J_HOME}` or by extracting the tar with `--same-owner` the permissions are set as the package.

I would also argue that the docker-entrypoint should not change permissions at all. The base image is correct. If people mount directories they should be able to control them.

Cheers,
Ron

moxious:
Related - users have previously reported some permission denied errors in various combinations, when a host volume is mapped to /data -- in these cases, the docker container is trying to change permissions on a host-owned directory, which may be owned/controlled by a user other than the user running the docker executable.

rweverwijk:
Extra on this: we would like to control the /data/dbms with kubernetes secrets. 

Binding a secret as directory makes it readonly. The recursive `chown` and `chmod` fail for this reason.

jennyowen:
Permissions on mounted folders are a reoccurring problem with our docker images (and docker in general).
By default, Neo4j runs as the user `neo4j` (because https://github.com/neo4j/docker-neo4j/issues/50), who only exists in the container, not on the host. That means that it's hard to set up folders on the host, which this new user has permissions for.

This is quite a complicated issue, and it's hard to fix file permissions in a way that won't also break backwards compatibility for other users.

There's some discussion of that here:
https://github.com/neo4j/docker-neo4j/pull/161

You can use the `--user` flag on docker run to specify which user to run neo4j as. If you use a user that has write access to your data/log/conf folders then folder ownership is not changed. Note that the `--user` flag to docker is the user the container will run stuff as **inside the container** (see: https://docs.docker.com/engine/api/v1.30/#operation/ContainerCreate). This means if you specify a host user name, docker won't recognise it as having write access to your folder. You need to specify the **user ID**. As an example, you can do `--user="$(id -u)":"$(id -g)"` to specify the current user. 
This is a docker limitation and there's not much we can do about it.

Another alternative is to mount docker volumes (instead of host folders) to your container. For example:
```
docker volume create --name neo4jdata
docker volume create --name neo4jlogs
docker run \
[...]

PR #165, opened 2019-03-21
Docker entrypoint should not chmod or chown folders

rweverwijk:
This PR should solve permission issues with mounted folders.
The idea is that the provider of the mount is responsible for correct permissions for the docker image.

The entrypoint should check the correctness of the used mounts and give correct pointers to the user to help if there are missing permissions.

This PR should solve #164

### DIFF
--- a/src/3.5/Dockerfile
+++ b/src/3.5/Dockerfile
@@ -22,14 +22,13 @@ RUN apk add --no-cache --quiet \
     && rm ${NEO4J_TARBALL} \
     && mv "${NEO4J_HOME}"/data /data \
     && chown -R neo4j:neo4j /data \
-    && chmod -R 777 /data \
     && mv "${NEO4J_HOME}"/logs /logs \
     && chown -R neo4j:neo4j /logs \
-    && chmod -R 777 /logs \
     && chown -R neo4j:neo4j "${NEO4J_HOME}" \
-    && chmod -R 777 "${NEO4J_HOME}" \
     && ln -s /data "${NEO4J_HOME}"/data \
     && ln -s /logs "${NEO4J_HOME}"/logs \
+    && for _dir in "/data /logs $NEO4J_HOME";do find ${_dir} -type d -exec chmod 750 {} ";";done \
+    && for _dir in "/data /logs";do find ${_dir} -type f -exec chmod 640 {} ";";done \
     && apk del curl
 
 ENV PATH "${NEO4J_HOME}"/bin:$PATH
--- a/src/3.5/Dockerfile
+++ b/src/3.5/Dockerfile
@@ -25,9 +25,10 @@ RUN apk add --no-cache --quiet \
     && mv "${NEO4J_HOME}"/logs /logs \
     && chown -R neo4j:neo4j /logs \
     && chown -R neo4j:neo4j "${NEO4J_HOME}" \
+    && chmod -R 777 "${NEO4J_HOME}" \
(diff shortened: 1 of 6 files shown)


jennyowen:
Hi Ron, sorry for the wait. I was on vacation, but now I'm on the case!

The 3.5 tests failed with the error:
`docker: Error response from daemon: OCI runtime create failed: container_linux.go:344: starting container process caused "chdir to cwd (\"/var/lib/neo4j\") set in config.json failed: permission denied": unknown.`
`ERROR (test-starts-up-with-data-volume-and-user): trap on error (rc=126) near line 104`

I assume "line 104" refers to this, where the container is started:
https://github.com/neo4j/docker-neo4j/blob/master/test/helpers.sh#L104

Did you get that error when you run the tests locally? I realise we have no information about how to actually run these tests, so for 3.5 it's:

```
export NEO4J_VERSION=3.5.3
make clean test
```

rweverwijk:
Hi Jenny,

I didn't know how to run them, I will take a look at them tomorrow night.

Cheers,
Ron

rweverwijk:
small update from my side. Had some trouble running the test as it didn't work on OSX out of the box . Needed a test env with GNU tooling.

Setup is now correct I thing so will take a look at the issues this week.

rweverwijk:
hi @jennyowen,

At can you take a look again? tests run successfully on my machine, can't see the log of the failing build in TeamCity.

Cheers,
Ron

jennyowen:
@rweverwijk a couple of things.

1) Before I can accept the PR I'll need you to agree to our Contributor License Agreement here:
https://neo4j.com/developer/cla/
It's a one page document, and agreement just requires sending an email. Details are in the link.

2) It's very important we don't break functionality for existing users. Requiring users to `chmod` the mount folders before running the neo4j container would do that.
The promised behaviour for the current containers is that this should **just work**:
```bash
docker run \
    --volume=$HOME/neo4j/data:/data \
    --volume=$HOME/neo4j/logs:/logs \
    neo4j:3.5
```
We can't break that functionality in a patch, it has to wait for a new major/minor release.
What we can (and should) be doing is checking that the user can read and write to the mounted folders, and doing a `chmod` **only** if they cannot. There should also be better advice about how to fix permissions error when they arise. 

The current behaviour of automatically editing the folder permissions is definitely wrong, and your PR improves the error messaging and the permission checking immensely. However, I'll still need to make some modification so that it's compatible with the current workflow. Hope that's OK?

rweverwijk:
Hi Jenny.

1) I'm already in the system since Match 4th 2016
2) I'm okay with modifications, you can continue with this PR, use this as inspiration or completely ignore and close this, all fine by me and up to you ;)

jennyowen:
Hi Ron, excellent!
I've adapted your changes and fixed some bugs I found.
The container will still change the owner on mounted folders if the `--user` flag isn't supplied, but it should no longer `chmod` everything to allow all users to read/write.

For your use case, I'd recommend supplying the user and group IDs to `docker run` (if you weren't already). That's the docker recommended way of mounting host folders, since the process and folder will be owned by the same user.
```
docker run \
   -p 7474:7474 -p 7687:7687 \
   --user "$(id -u):$(id -g)" \
   -v /etc/passwd:/etc/passwd:ro \
   -v OTHERMOUNTS... \
    neo4j:latest
```

The changes should be in the following patches onwards:
- 3.3.10
- 3.4.14
- 3.5.6

PR #168, opened 2019-05-16
Switch to AdoptOpenJDK-based JVM base image

jakewins:
The background here is that - again - the Debian OpenJDK builds turn out
to have the wrong version numbers on them, with the package being
stamped as a java update that was released after the package was built.

See https://mail.openjdk.java.net/pipermail/jdk8u-dev/2019-May/009330.html

We don't use the Debian packages, we use Alpine, but inspecting the
provenance of the Alpine OpenJDK packaging, it seems to suffer from the same
issue. The current latest alpine OpenJDK has versioning that does not
seem to line up. At the very least, the source of those builds is of a
similar kind to Debian.

This moves our base image to images provided by AdoptOpenJDK, a well
regarded source of up-to-date OpenJDK builds. The base image remains
Alpine Linux.

### DIFF
--- a/src/3.1/Dockerfile
+++ b/src/3.1/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8-jre-alpine
+FROM adoptopenjdk/openjdk8:alpine-jre
 
 RUN addgroup -S neo4j && adduser -S -H -h /var/lib/neo4j -G neo4j neo4j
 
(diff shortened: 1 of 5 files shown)


jakewins:
Note that I'm not changing below Neo4j 3.1; those images are not alpine based.. and only 3.4 and 3.5 are officially supported, so I don't think that should be a problem

PR #169, opened 2019-05-25
Check groups when determining if a directory is writable

eastlondoner:
this needs to be copy pasted to all the 3.x folders and test coverage must be added.
I leave both as an exercise for the reader. 👶 🍼 👪 etc.


jennyowen:
looks good to me, and the current tests still pass. Good job!

PR #181, opened 2019-07-05
Adding the ability to install neo4j labs plugins on container startup

eastlondoner:
This time with Java tests.

To test the functionality I defined an internal testing plugin - for which the expected versions.json is hosted on an internal testcontainers url. 

At test time I create the plugin jar using the relevant neo4j version and host it on a port that's exposed to the test container. This way the fetching-at-runtime logic is excercised but I don't introduce a dependency on any "real" neo4j plugins.

### DIFF
--- a/src/test/java/com/neo4j/docker/TestPluginInstallation.java
+++ b/src/test/java/com/neo4j/docker/TestPluginInstallation.java
@@ -1,0 +1,141 @@
+package com.neo4j.docker;
+
+import com.neo4j.docker.plugins.ExampleNeo4jPlugin;
+import com.neo4j.docker.utils.HostFileHttpHandler;
+import com.neo4j.docker.utils.HttpServerRule;
+import com.neo4j.docker.plugins.JarBuilder;
+import com.neo4j.docker.utils.SetContainerUser;
+import com.neo4j.docker.utils.TestSettings;
+import org.junit.Rule;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.migrationsupport.rules.EnableRuleMigrationSupport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.Testcontainers;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.shaded.com.google.common.io.Files;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+
+import org.neo4j.driver.AuthTokens;
+import org.neo4j.driver.Driver;
+import org.neo4j.driver.GraphDatabase;
+import org.neo4j.driver.Record;
+import org.neo4j.driver.Session;
+import org.neo4j.driver.StatementResult;
+
+import static com.neo4j.docker.utils.TestSettings.NEO4J_VERSION;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@EnableRuleMigrationSupport
+public class TestPluginInstallation
+{
+    private static final int DEFAULT_BROWSER_PORT = 7474;
+    private static final int DEFAULT_BOLT_PORT = 7687;
+
+    private static final String versions = "versions.json";
+    private static final String myPlugin = "myPlugin.jar";
+
+    private static final Logger log = LoggerFactory.getLogger( TestPluginInstallation.class );
+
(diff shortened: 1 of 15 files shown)


eastlondoner:
I've had to use`env.NEO4JVERSION` in the pom because that is the env var being set in TC but I believe it _ought_ to be `env.NEO4J_VERSION` because that is what we use in devenv and the Makefile.
See this card: https://trello.com/c/xOCjuAaA/250-team-city-configuration-mistake-neo4jversion-should-be-neo4jversion

jennyowen:
@eastlondoner regarding the `NEO4J_VERSION` vs `NEO4JVERSION`. It's been bugging me for a while that our docker build uses `NEO4J_VERSION` but `NEO4JVERSION` is used literally everywhere else in our tools.

eastlondoner:
Argh! You’re right on both counts.
Bear with, I’ll add the test changes as soon as I have WiFi.

src/test/java/com/neo4j/docker/TestPluginInstallation.java
```diff
@@ -0,0 +1,140 @@
...
+    private static final int DEFAULT_BROWSER_PORT = 7474;
+    private static final int DEFAULT_BOLT_PORT = 7687;
+
+    private static final String versions = "/versions.json";
+    private static final String myPlugin = "/myPlugin.jar";
+
+    private static final Logger log = LoggerFactory.getLogger( TestPluginInstallation.class );
+
+    @Rule
+    public TemporaryFolder plugins = new TemporaryFolder();
```
jennyowen: There's a variable `TestSettings.TEST_TMP_FOLDER` which I've been making the tests write temp folders under. It'd be good if this test used that too.

jennyowen: There's a variable `TestSettings.TEST_TMP_FOLDER` which I've been making the tests write temp folders under. It'd be good if this test used that too.

src/test/java/com/neo4j/docker/TestPluginInstallation.java
```diff
@@ -0,0 +1,140 @@
...
+{
+    private static final int DEFAULT_BROWSER_PORT = 7474;
+    private static final int DEFAULT_BOLT_PORT = 7687;
+
+    private static final String versions = "/versions.json";
+    private static final String myPlugin = "/myPlugin.jar";
+
+    private static final Logger log = LoggerFactory.getLogger( TestPluginInstallation.class );
+
+    @Rule
```
jennyowen: In JUnit5 TemporaryFolder is replaced by:
https://junit.org/junit5/docs/current/user-guide/#writing-tests-built-in-extensions-TempDirectory

src/test/java/com/neo4j/docker/TestPluginInstallation.java
```diff
@@ -0,0 +1,140 @@
...
+import com.neo4j.docker.utils.HostFileHttpHandler;
+import com.neo4j.docker.utils.HttpServerRule;
+import com.neo4j.docker.utils.JarBuilder;
+import com.neo4j.docker.utils.SetContainerUser;
+import com.neo4j.docker.utils.TestSettings;
+import org.junit.Rule;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.migrationsupport.rules.EnableRuleMigrationSupport;
```
jennyowen: You're using a mix of JUnit4 and JUnit5 here. We've been using JUnit5 in the rest of the tests.
JUnit4 Rules are better imho, I don't know how you want to do it here. But you should be consistent at least.

jennyowen: If you decide to keep this JUnit4 then we should migrate the rest of the tests back to junit4. But that can be a seperate PR. 
I think there's migration packages you'll need to add to the pom if you keep it junit4:
https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4-running

eastlondoner: I've added the migration package so that I can still use the `ExternalResource` `Rule` for the HTTP Server

src/test/java/com/neo4j/docker/utils/HostFileHttpHandler.java
```diff
@@ -0,0 +1,33 @@
+package com.neo4j.docker.utils;
```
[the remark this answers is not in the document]
eastlondoner: I've moved the two totally plugin-specific things (the Plugin class and JarBuilder) to `com.neo4j.docker.plugins`

eastlondoner:
image

src/test/java/com/neo4j/docker/utils/HostFileHttpHandler.java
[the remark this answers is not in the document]
jennyowen: they still declare  `package com.neo4j.docker.utils` at the top though

eastlondoner:
packages sorted out

src/test/java/com/neo4j/docker/utils/HostFileHttpHandler.java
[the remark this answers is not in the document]
eastlondoner: doh

PR #183, opened 2019-07-22
Docker entrypoint fix

bfeshti:
Moved default config values from the docker-entrypoint.sh to neo4j.conf.
Fixed overriding of the configurations set by users on neo4j.conf from docker-entrypoint.sh.

### DIFF
--- a/docker-image-src/3.2/docker-entrypoint.sh
+++ b/docker-image-src/3.2/docker-entrypoint.sh
@@ -122,7 +122,7 @@ fi
 if [ -d /import ]; then
     NEO4J_dbms_directories_import="/import"
 fi
-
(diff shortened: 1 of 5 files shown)


jennyowen:
1) Don't give your commits the same message each time. The message should describe the changes in that specific commit. Otherwise when someone looks back to try and find what your code changes were, it's really hard to find. This is something I should do better at too. 
See: https://chris.beams.io/posts/git-commit/

2) Where are the tests for the new feature? **Never ever** submit behaviour changes or new features without making sure there are tests which exercise the new/changed code or functionality.

I have many other questions, but maybe you should walk me through what this code is doing and why.

PR #186, opened 2019-08-06
Fixed ovverriding of custom configurations by docker-entrypoint

bfeshti:
* Added a function in `docker-entrypoint.sh` to pass the default configs to the `neo4j.conf` file only when they are not set manually by the user.
* Added 2 Tests to test if the configs are passed correctly into `neo4j.conf`.
* Removed some commented lines from the `neo4j.conf`file used for the tests.


src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -135,4 +136,53 @@ void testReadTheConfFile() throws Exception
...
+        SetContainerUser.currentlyRunningUser( container );
+        //Create neo4j.conf file
+        Path confFile = Paths.get( "src", "test", "resources", "neo4j.conf" );
+        Files.copy(confFile, confMount.resolve("neo4j.conf"));
+        //Start the container
+        container.setWaitStrategy( Wait.forHttp( "/" ).forPort( 7474 ).forStatusCode( 200 ) );
+        container.start();
+        //Read the config file to check if the config is not overriden
+        Stream<String> lines = Files.lines(confMount.resolve("neo4j.conf"));
+        Optional<String> boltListenAdressMatch = lines.filter(s -> s.contains("dbms.connector.https.listen_address=:8483")).findFirst();
```
jennyowen: minor variable name quibble, but this is actually the https listen address not the bolt listen address.

src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -135,4 +136,53 @@ void testReadTheConfFile() throws Exception
...
+        Path confMount = HostFileSystemOperations.createTempFolderAndMountAsVolume( container, "conf-", "/conf" );
+        SetContainerUser.currentlyRunningUser( container );
+        //Create neo4j.conf file
+        Path confFile = Paths.get( "src", "test", "resources", "neo4j.conf" );
+        Files.copy(confFile, confMount.resolve("neo4j.conf"));
+        //Start the container
+        container.setWaitStrategy( Wait.forHttp( "/" ).forPort( 7474 ).forStatusCode( 200 ) );
+        container.start();
+        //Read the config file to check if the config is not overriden
+        Stream<String> lines = Files.lines(confMount.resolve("neo4j.conf"));
```
jennyowen: This test is copy a test config file to the location at `confMount.resolve("neo4j.conf")` then starting neo4j.
Since we're starting neo4j in this test rather than doing `dump-config`, that file is never overridden. So what we're actually testing here is that the test conf we copied over has a configuration line that we know we set. Does that make sense?

There are 2 solutions:
1) do a `dump-config` command instead of starting neo4j like `testCommentedConfigsAreReplacedByDefaultOnes()` does.
2) check the log output from neo4j like the test `testReadTheConfFile()` does.

(I think option 1 is better. Since we don't really need to be starting neo4j, it'll be much quicker).

src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -135,4 +136,53 @@ void testReadTheConfFile() throws Exception
         //Kill the container
         container.stop();
     }
+
```
jennyowen: We have a test `testAdminConfOverride()` which checks that configs from environment variables override default settings. 
Could you add a test that checks configs from environment variables override configs from the neo4j.conf file?

It'd be pretty similar to the test `testConfigsAreNotOverridenByDockerentrypoint()` only you start the container with an environment variable for the same config as you set in the test conf file. Verify that the value used is the same as the environment variable and NOT the default and NOT the value from the conf.

src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -135,4 +136,53 @@ void testReadTheConfFile() throws Exception
...
+        SetContainerUser.currentlyRunningUser( container );
+        //Create neo4j.conf file
+        Path confFile = Paths.get( "src", "test", "resources", "neo4j.conf" );
+        Files.copy(confFile, confMount.resolve("neo4j.conf"));
+        //Start the container
+        container.setWaitStrategy( Wait.forLogMessage( ".*Config Dumped*." , 1 ).withStartupTimeout( Duration.ofSeconds( 10 ) ) );
+        container.setCommand("dump-config");
+        container.start();
+        //Read the config file to check if the config is appended correctly
+        Stream<String> lines = Files.lines(confMount.resolve("neo4j.conf"));
```
jennyowen: The `Files.lines` code was good for parsing the `debug.log`, but there's the method `parseConfFile()` which could be used for parsing the dumped conf file. It's easier to step through with the debugger to see all the configs that are actually being set.

src/test/java/com/neo4j/docker/TestConfSettings.java
jennyowen: again, since we're parsing a conf file you could use `parseConfFile()` to get a map of all the configuration names and their values.

jennyowen:
I don't know why the dumpConfig test started failing suddenly but I can guess why it did.
Could you change the line where it's setting the wait strategy to `null` here: https://github.com/neo4j/docker-neo4j/blob/47fb7b63fa813390e1f44b66bc390fb6b6e51882/src/test/java/com/neo4j/docker/TestMounting.java#L99

to the same wait strategy you used here:
https://github.com/neo4j/docker-neo4j/blob/47fb7b63fa813390e1f44b66bc390fb6b6e51882/src/test/java/com/neo4j/docker/TestConfSettings.java#L152

I think that should fix it.

src/test/java/com/neo4j/docker/TestConfSettings.java
jennyowen: this comment still applies

src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -16,123 +16,187 @@
...
-        {
-            String[] params = scanner.nextLine().split( "=", 2 );
-            log.debug( params[0] +"\t:\t"+ params[1] );
-            configurations.put( params[0], params[1] );
+    private Map<String, String> parseConfFile(File conf) throws FileNotFoundException {
+        Map<String, String> configurations = new HashMap<>();
+        Scanner scanner = new Scanner(conf);
+        while (scanner.hasNextLine()) {
+            String[] params = scanner.nextLine().split("=", 2);
+            log.debug(params[0] + "\t:\t" + params[1]);
```
jennyowen: `testAdminConfOverride()` failed here when I ran it.
Change this while loop to:
```
while ( scanner.hasNextLine() )
        {
            String[] params = scanner.nextLine().split( "=", 2 );
            if(params.length < 2)
            {
                continue;
            }
            log.debug( params[0] + "\t:\t" + params[1] );
            configurations.put( params[0], params[1] );
        }
```

PR #189, opened 2019-08-09
can now set non-root user via envoriment variable.

jennyowen:
This is to enable the tests to run inside a docker container, even though they themselves spawn new containers.
It was failing in our CI pipeline because of permission errors on mounted folders when testing mounting as non-root users.


jennyowen:
I should add that this change only applies to the tests, the behaviour of the neo4j image is unaffected.

PR #192, opened 2019-09-11
docker-entrypoint will now set default configs based on Neo4j Edition

bfeshti:
* Added the ability to set default configs based on Neo4j Edition.
* Added a test to check if the right configs are set correctly in the Enterprise edition.
* Added a test to check that there are no Enterprise edition configs set to the Community edition.

### DIFF
--- a/src/test/java/com/neo4j/docker/TestConfSettings.java
+++ b/src/test/java/com/neo4j/docker/TestConfSettings.java
@@ -208,31 +208,27 @@ public class TestConfSettings {
     }
 
     @Test
-    void testEnterpriseConfigsAreSetCorrectly() throws Exception {
+    void testEnterpriseOnlyDefaultsConfigsAreSet () throws Exception {
         Assumptions.assumeTrue(TestSettings.EDITION == TestSettings.Edition.ENTERPRISE,
                 "This is testing only ENTERPRISE EDITION configs");
         //Create container
         GenericContainer container = createContainer();
-        //Mount /conf
-        Path confMount = HostFileSystemOperations.createTempFolderAndMountAsVolume(container, "conf-", "/conf");
+        //Mount /logs
         Path logMount = HostFileSystemOperations.createTempFolderAndMountAsVolume(container, "logs-", "/logs");
         SetContainerUser.nonRootUser(container);
-        //Create EnterpriseCOnf.conf file
-        Path confFile = Paths.get("src", "test", "resources", "confs", "EnterpriseConf.conf");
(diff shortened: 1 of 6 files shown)


src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -199,12 +198,57 @@ void testEnvVarsOverride() throws Exception {
...
         Stream<String> lines = Files.lines(logMount.resolve("debug.log"));
         Optional<String> heapSizeMatch = lines.filter(s -> s.contains("dbms.memory.pagecache.size=512m")).findFirst();
         lines.close();
         Assertions.assertTrue(heapSizeMatch.isPresent(), "dbms.memory.pagecache.size was not set correctly");
         //Kill the container
         container.stop();
     }
+
+    @Test
+    void testEnterpriseConfigsAreSetCorrectly() throws Exception {
```
jennyowen: The test name doesn't really describe what the test is checking. When I/you look back on this in 2 or 3 months time we'd have to read all the code to know what is actually being tested. Can you name it something like testEnterpriseOnlyDefaultsConfigsAreSet or something similar?

src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -199,12 +198,53 @@ void testEnvVarsOverride() throws Exception {
...
+        GenericContainer container = createContainer();
+        //Mount /logs
+        Path logMount = HostFileSystemOperations.createTempFolderAndMountAsVolume(container, "logs-", "/logs");
+        SetContainerUser.nonRootUser(container);
+        //Start the container
+        container.setWaitStrategy(Wait.forHttp("/").forPort(7474).forStatusCode(200));
+        container.start();
+        //Read debug.log to check that causal_clustering confs are set successfully
+        Stream<String> lines = Files.lines(logMount.resolve("debug.log"));
+        Optional<String> ccPresent = lines.filter(s -> s.contains("causal_clustering.transaction_advertised_address")).findFirst();
```
jennyowen: I think it would be better if we actually tested that the configuration is also set to the value we expect. It looks as though entrypoint resolves `$(hostname)` to be the first 12 digits of the container id, so this works for me:

```
        String expectedTxAddress = container.getContainerId().substring( 0, 12 ) + ":6000";
        Stream<String> lines = Files.lines(logMount.resolve("debug.log"));
        Optional<String> ccPresent = lines.filter(s -> s.contains("causal_clustering.transaction_advertised_address="+expectedTxAddress)).findFirst();
        lines.close();
```

PR #205, opened 2019-11-15
fixed jvm.additional overriden by docker-entrypoint

bfeshti:
* Fixed the issue where the custom values set by users for `dbms.jvm.additional` were being overriden by `docker-entrypoint.sh`.
* Added a small test for it.


bfeshti:
I think we have the same problem with it also,I will fix it. Thanks.

bfeshti:
Found a new more reliable way to fix it that also includes the ```NEO4J_wrapper_java_additional```

PR #206, opened 2019-11-15
Add neo4j m2 repo (community) & always run the plugins tests

eastlondoner:


pom.xml
```diff
@@ -13,6 +13,17 @@
...
 	</properties>
 
+	<repositories>
+		<repository>
+			<snapshots>
+				<enabled>false</enabled>
+			</snapshots>
+			<id>neo4j-community</id>
+			<name>neo4j-community</name>
+			<url>http://m2.neo4j.com/community</url>
```
eastlondoner: @glindroth / @jennyowen can you confirm that this is the correct way to get pre-release jars?

jennyowen: @eastlondoner no, m2.neo4j.org got closed down and we shouldn't use it any more. The community maven artifacts are here: https://bintray.com/neo4j/community/public

The site should have maven instructions but it's something like: 

```
                <repository>
                    <snapshots>
                        <enabled>false</enabled>
                    </snapshots>
                    <id>bintray-neo4j-community</id>
                    <name>bintray</name>
                    <url>https://neo4j.bintray.com/community</url>
                </repository>
```

I'm not sure we want to be introducing more very specific maven dependencies to the tests though? What do you need this for? Or is this PR so you can investigate the plugin test failure and the pom change isn't going to be in the final merge?

Makefile
```diff
@@ -21,11 +21,17 @@ series := $(shell echo "$(NEO4JVERSION)" | sed -E 's/^([0-9]+\.[0-9]+)\..*/\1/')
...
+test: test-enterprise test-community
 .PHONY: test
 
+test-enterprise: tmp/.image-id-enterprise
+> mvn test -Dimage=$$(cat $<) -Dedition=enterprise -Dversion=$(NEO4JVERSION)
+.PHONY: test-enterprise
+
+test-community: tmp/.image-id-community
+> mvn test -Dimage=$$(cat $<) -Dedition=community -Dversion=$(NEO4JVERSION)
+.PHONY: test-commercial
```
jennyowen: `community` right?

eastlondoner: thanks

pom.xml
eastlondoner: @jennyowen I haven't introduced and new dependencies.

Neo4j has decided not to publish pre release versions of neo4j jars _that this project has dependencies on_ to Maven Central. This is a consequence of that.

This change is required to make this repository function if NEO4JVERSION is set to a version that is present on the community binary and not in Maven Central.

eastlondoner:
Please see the related trello card.
I've tidied up the commit history and added an environment variable so we can enable/disable the test in a more fine grained way using team city environment parameters

src/test/java/com/neo4j/docker/TestPluginInstallation.java
```diff
@@ -60,13 +61,7 @@
     @BeforeAll
     public static void checkVersionIsCompatibleWithTest()
```
jennyowen: do we still need this method?

eastlondoner: hopefully not!
I'll tidy it up once this is definitely sorted out

PR #210, opened 2019-11-25
Try with resources

jennyowen:
TestContainers doesn't handle closing an already terminated container very well, and NOT closing the containers after each test leads to stray containers running and consuming memory, leading to test failures.
I've changed the unit tests to take advantage of the container's autoclosable ability, so that the containers is automatically closed after the try with resources goes out of scope.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,6 @@
 /target/
 
 # simlinks to dev scripts in https://github.com/neo-technology/teamcity-witchcraft
-DownloadTool.py
+download_tool.py
 docker_tests_get_installers.py
 __pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,8 @@
 /devenv.local
 /local-mounts/
 /target/
+
+# simlinks to dev scripts in https://github.com/neo-technology/teamcity-witchcraft
+DownloadTool.py
+docker_tests_get_installers.py
+__pycache__/
--- a/src/test/java/com/neo4j/docker/TestBasic.java
+++ b/src/test/java/com/neo4j/docker/TestBasic.java
@@ -1,6 +1,5 @@
 package com.neo4j.docker;
 
-import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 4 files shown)


bsideup:
> TestContainers doesn't handle closing an already terminated container very well

Could you please share more details on this please?
Also, `close()` simply delegates to `stop()`, which makes me think that if you see that `try-with-resources` gives better result then there may be some issue with how you integrate Testcontainers into your test suite. Eager to help finding what was causing the close handling not behave well :)

bsideup:
(Although I do prefer using `try-with-resources` in scenarios like yours and the PR is still a good improvement! 👍)

jennyowen:
@bsideup 
I suppose by
> TestContainers doesn't handle closing an already terminated container very well

I really mean "doesn't handle closing an already terminated container **in the way I need here**"

There are quite a few tests where the container is expected to fail to start. In this case, the container dies but we still have the object reference in java.
For example: https://github.com/neo4j/docker-neo4j/blob/54d1bd4d4e63508309ffeaf90561c47157be67c2/src/test/java/com/neo4j/docker/TestBasic.java#L80
(Admittedly our test code here is actually pretty bad and inconsistent).
Since this is a test, there's the chance that the test failed and the container didn't close itself, so we have to make sure the container is closed in code. Doing try-with-resources won't fix these kinds of race conditions though.

It would be very nice if calling `container.close()` on an already closed container didn't fail. That might be too much of a behaviour change though? It would definitely be handy if there was an inbuilt `WaitStrategy` that would wait until the container closes itself.

bsideup:
> It would be very nice if calling container.close() on an already closed container didn't fail

Could you please share the exception? It really should not fail

jennyowen:
@bsideup 
The stack trace I have is:
```
java.lang.reflect.UndeclaredThrowableException
	at com.neo4j.docker.TestBasic.killContainer(TestBasic.java:33)
Caused by: java.lang.reflect.InvocationTargetException
	at com.neo4j.docker.TestBasic.killContainer(TestBasic.java:33)
Caused by: com.github.dockerjava.api.exception.ConflictException: 
{"message":"Cannot kill container: bfb40a1e621564d4af8a4f56c7eb2dece42227a6c1ea99354b26d159b77a6f06: Container bfb40a1e621564d4af8a4f56c7eb2dece42227a6c1ea99354b26d159b77a6f06 is not running"}
```
TestBasic.java:33 is calling `stop()` on a container.
It doesn't happen very often, and only occurs on tests where the container dies during the test before `stop` is called. This exception was from this test: https://github.com/neo4j/docker-neo4j/blob/54d1bd4d4e63508309ffeaf90561c47157be67c2/src/test/java/com/neo4j/docker/TestBasic.java#L80

bsideup:
@jennyowen 

Thanks! Could you please [report it in Testcontainers](https://github.com/testcontainers/testcontainers-java/issues/new)? We will try to fix it :)

jennyowen:
@bsideup ok done, thanks!
https://github.com/testcontainers/testcontainers-java/issues/2102
I tried getting better reproduction steps but couldn't.

PR #217, opened 2019-12-11
apply default configurations for plugins when installing a plugin

eastlondoner:
(if the user has not already set the configuration value in question)

n.b. requested on community slack

also improves behaviour when no jar is found (exits after error rather than proceeding on to a curl/wget that will fail)


eastlondoner:
This is almost exactly the same as what neo4j desktop does

PR #234, opened 2020-03-04
Remove udc.source from 4.0.x

bfeshti:
Since udc.source is deprecated starting from neo4j 4.0 there is no sense into keeping it in the docker image.


src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -307,12 +307,13 @@ void testJvmAdditionalNotOverridden() throws Exception
         }
 
         //Read the debug.log to check that dbms.jvm.additional was set correctly
+        Assumptions.assumeFalse( TestSettings.NEO4J_VERSION.isAtLeastVersion( Neo4jVersion.NEO4J_VERSION_400), "test not applicable in versions newer than 4.0." );
```
jennyowen: If you do test assumptions you should put them to the top of the test. They're for ignoring the test if it's running in invalid conditions. You don't want to be doing slow/risky/expensive operations (here the slow and memory intensive operation of starting a docker container and creating mounted folders which neo4j fills with junk) only to **then** check whether you're even testing something you want to test.

Issue #240, opened 2020-03-20
Can't connect to DB using `latest` image: SessionExpired: WebSocket connection failure. Due to security constraints in your web browser, the reason for the failure is not available to this Neo4j Driver


TravelingTechGuy:
# Environment

- Windows 10
- Docker Desktop (version 2.2.0.4 Engine 19.3.8 Compose 1.24.4)
- Browser: Brave

# Steps to reproduce

1. Get latest image: `docker pull neo4j`
1. Run:
```
docker run \
    --publish=7474:7474 --publish=7687:7687 \
    --volume=$HOME/neo4j/data:/data \
    --env=NEO4J_AUTH=none \
    --env SECURE_FILE_PERMISSIONS=yes \
    --name my_neo4j \
    neo4j
```
1. Connect to browser at localhost:7474
1. Regardless of which connection method I use ("no authentication" or user/pass) I get the same error message: `SessionExpired: WebSocket connection failure. Due to security constraints in your web browser, the reason for the failure is not available to this Neo4j Driver`

I tried following a much earlier report to uncomment a line in the conf file, but cannot get VI to work inside the container shell.

What am I missing?

jennyowen:
@TravelingTechGuy The docker run command you provided should work without needing to modify the conf (assuming the database in `$HOME/neo4j/data` isn't broken somehow).

I have a feeling that error could be caused by a browser setting.
Could you start a neo4j container without mounting anything:
```
docker run \
    --publish=7474:7474 --publish=7687:7687 \
    --env=NEO4J_AUTH=none \
    neo4j
```
Then see if `cypher-shell` can connect to it:

`docker run  --network host neo4j cypher-shell "CALL dbms.procedures;"`

That should test whether the error is coming from the system browser.

TravelingTechGuy:
That seems to work @jennyowen! 
But I've used this browser with previous images, and would like yo use it with Neo4J 4. What do I need to do in order to get the browser to work with this image?

TravelingTechGuy:
Can now report getting the same error in FireFox and Edge (classic, not Chromium based).
Full error message (since the one of the screen is truncated):

```
[...]

PR #241, opened 2020-03-25
removed overrides incompatible with windows networking

jennyowen:
Removing several unnecessary advertised and listen address overrides from the entrypoint.
In Linux `0.0.0.0` automatically resolves to `localhost`, but this is not true in Windows.
Removing the uneccessary overrides from the docker configuration should make docker more closely match the non-docker settings.

This should fix issue #240

### DIFF
--- a/docker-image-src/4.0/docker-entrypoint.sh
+++ b/docker-image-src/4.0/docker-entrypoint.sh
@@ -403,7 +403,7 @@ if [ "${cmd}" == "neo4j" ]; then
 fi
 
 declare -A COMMUNITY
-# declare -A ENTERPRISE
+declare -A ENTERPRISE
 
 COMMUNITY=(
      [dbms.tx_log.rotation.retention_policy]="100M size"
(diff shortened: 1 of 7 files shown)


docker-image-src/4.0/docker-entrypoint.sh
```diff
@@ -403,21 +403,12 @@ if [ "${cmd}" == "neo4j" ]; then
...
 COMMUNITY=(
      [dbms.tx_log.rotation.retention_policy]="100M size"
      [dbms.memory.pagecache.size]="512M"
      [dbms.default_listen_address]="0.0.0.0"
-     [dbms.connector.https.advertised_address]="0.0.0.0:7473"
-     [dbms.connector.http.advertised_address]="0.0.0.0:7474"
-     [dbms.connector.bolt.advertised_address]="0.0.0.0:7687"
-)
-
-ENTERPRISE=(
```
hugofirth: How do you feel about keeping the Enterprise array, even if its empty, just to indicate to future authors *how* you should add Enterprise overrides?

docker-image-src/4.0/docker-entrypoint.sh
```diff
@@ -428,16 +419,16 @@ for conf in ${!COMMUNITY[@]} ; do
...
-
-    if [ "${NEO4J_EDITION}" == "enterprise" ];
-    then
-       if ! grep -q "^$conf" "${NEO4J_HOME}"/conf/neo4j.conf
-       then
-        echo -e "\n"$conf=${ENTERPRISE[$conf]} >> "${NEO4J_HOME}"/conf/neo4j.conf
-       fi
-    fi
-done
+#for conf in ${!ENTERPRISE[@]} ; do
```
hugofirth: Same as the above - if the array is empty then maybe we could keep this uncommented, as its a no-op?

src/test/java/com/neo4j/docker/TestConfSettings.java
```diff
@@ -264,27 +264,28 @@ void testEnterpriseOnlyDefaultsConfigsAreSet () throws Exception
...
-        {
-            //Mount /logs
-            Path logMount = HostFileSystemOperations.createTempFolderAndMountAsVolume( container, "logs-enterprisesettingsnotincommunity-", "/logs" );
-            debugLog = logMount.resolve( "debug.log" );
-            SetContainerUser.nonRootUser( container );
-            //Start the container
-            container.setWaitStrategy( Wait.forHttp( "/" ).forPort( 7474 ).forStatusCode( 200 ) );
-            container.start();
-        }
+//    @Test
```
hugofirth: Is this test no longer valid?

jennyowen: the test checks that in community edition we aren't adding any enterprise specific configurations (like for clustering). It used to be that they were both the same, and community edition would always have warnings about invalid settings. 

Anyway, now that there are no enterprise specific settings there's nothing to actually test for. So, yes the test is not valid any more, but it could become valid again in the future?

PR #251, opened 2020-05-11
Updated docker SSL behaviour for 4.0 onwards

jennyowen:
SSL instructions were not working for 4.0 so I've updated the behaviour of the `/ssl` mounted folder checks.
I also realised that if a folder is mounted to any of the expected mount points, we should be checking if it's readable before starting neo4j, otherwise it'll fail when trying to read. These checks had been wrapped with the unrelated `secure_mode_enabled` flag, which should only be used if we are going to do a `chown` on the mounted folder.

changelog: [4.0, 4.1, packaging, docker]


docker-image-src/4.0/docker-entrypoint.sh
```diff
@@ -206,6 +206,34 @@ function install_neo4j_labs_plugins
   rm "${_old_config}"
 }
 
+function process_ssl_policies
+{
+    local _found_a_policy_scope=false
+    local _ssl_policy_scopes=("bolt" "https" "cluster" "backup" "fabric")
+    for scope in ${_ssl_policy_scopes[@]}
+    do
+        echo $scope
```
eastlondoner, line 215: This echo should either be more explicit "checking if ssl policy for \"$scope\" is provided" or it should just be removed

src/test/resources/ssl/private.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
eastlondoner, line 1: I think it might be worth renaming this file something like `dummy_private.key` to make it clear that we haven't accidentally checked in some real private key.

Some security / vulnerability scanners might bork at this being in the repo as well. It's worth considering if you can create one on the fly (it doesn't need to be done in java - it could be done using `make`)

docker-image-src/4.0/Dockerfile
```diff
@@ -22,14 +22,14 @@ RUN apt update \
     && tar --extract --file ${NEO4J_TARBALL} --directory /var/lib \
     && mv /var/lib/neo4j-* "${NEO4J_HOME}" \
     && rm ${NEO4J_TARBALL} \
+    && chown -R neo4j:neo4j "${NEO4J_HOME}" \
+    && chmod -R 777 "${NEO4J_HOME}" \
     && mv "${NEO4J_HOME}"/data /data \
     && mv "${NEO4J_HOME}"/logs /logs \
     && chown -R neo4j:neo4j /data \
-    && chmod -R 777 /data \
     && chown -R neo4j:neo4j /logs \
```
eastlondoner, line 30: since you've moved the `chown -R` and `chmod -R` on `${NEO4JHOME}` up and it's recursive you should be able to get rid of the separate `chown` and `chmod` on `/data` and `/logs`

src/test/java/com/neo4j/docker/TestSSL.java
```diff
@@ -0,0 +1,165 @@
...
+		}
+	}
+
+	@Test
+	void testCanEncryptHttps() throws Exception
+	{
+		try(GenericContainer container = createBasicContainer())
+		{
+			Path sslFolder = createSSLDirectory( container, "https" );
+			HostFileSystemOperations.mountHostFolderAsVolume( container, sslFolder, "/var/lib/neo4j/certificates" );
```
eastlondoner, line 67: why are we mounting to `/var/lib/neo4j/..` here? can we have a comment to explain why we're doing this rather than mounting to `/ssl`?

eastlondoner:
If we're magically doing configuration we should also consider the following settings:
`dbms.connector.bolt.tls_level` -> should we switch this to REQUIRED
`dbms.connector.http.enabled` -> should we disable this if https is available

docker-image-src/4.0/docker-entrypoint.sh
```diff
@@ -206,6 +206,34 @@ function install_neo4j_labs_plugins
...
+        # does this policy folder exist in dir
+        if [[ -d /ssl/$scope ]]; then
+            local _connector_flag=NEO4J_dbms_connector_${scope}_enabled
+            : ${!_connector_flag:=true}
```
eastlondoner, line 219: according to the docs `dbms.connector.<foo>` is only available for `bolt` `http` and `https`.
If we do this for the others I presume we will get warnings logged about unrecognised config values

PR #255, opened 2020-05-21
improved readwrite checks on mounted folders

jennyowen:
I separated the rw checks from #251 into a new PR.
There is a question of what to do with the mounted `/ssl` volume since the configuration parameter that was being set is unused in 4.0 so I changed it to soft link `/ssl` to `$neo4j_home/certificates` instead. There's no equivalent to the old parameter in 4.0 onwards.

### DIFF
--- a/docker-image-src/4.1/docker-entrypoint.sh
+++ b/docker-image-src/4.1/docker-entrypoint.sh
@@ -95,7 +95,7 @@ function check_mounted_folder_readable
     fi
 }
 
-function check_mounted_folder_with_chown
+function check_mounted_folder_writable_with_chown
 {
 # The /data and /log directory are a bit different because they are very likely to be mounted by the user but not
 # necessarily writable.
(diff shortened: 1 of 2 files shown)


docker-image-src/4.1/docker-entrypoint.sh
```diff
@@ -308,56 +308,48 @@ unset NEO4J_dbms_txLog_rotation_retentionPolicy NEO4J_UDC_SOURCE \
...
 
 if [ -d /ssl ]; then
-    if secure_mode_enabled; then
-    	check_mounted_folder_readable "/ssl"
-    fi
-    : ${NEO4J_dbms_directories_certificates:="/ssl"}
+    check_mounted_folder_readable "/ssl"
+    ln -s /ssl "${NEO4J_HOME}"
+    rm -rf "${NEO4J_HOME}"/certificates
+    mv  "${NEO4J_HOME}"/ssl "${NEO4J_HOME}"/certificates
```
eastlondoner, line 319: how about
```
check_mounted_folder_readable "/ssl"
rm -rf "${NEO4J_HOME}"/certificates
ln -s /ssl/ "${NEO4J_HOME}/certificates/"
```

PR #263, opened 2020-06-16
Refactored code to generate test folders

jennyowen:
Now there's a method to specify if you want the test folder to sit under a specific parent folder, which makes grouping test folder outputs much easier


jennyowen:
@eastlondoner you can now use this refactored code in your PR!

PR #283, opened 2021-01-18
Add Bloom Plugin

GlacialBoy:
Allow Bloom to be installed using the "NEO4JLABS_PLUGINS".  
Bloom plugin requires a license file ( https://neo4j.com/docs/bloom-user-guide/current/bloom-installation/ ), in order to get the plugin running the user needs to share a volume like in the following example:
```
docker run -it --rm \
  -v $HOME/bloom.license:/licenses/bloom.license \
  --env NEO4J_AUTH=neo4j/test \
  --env=NEO4J_ACCEPT_LICENSE_AGREEMENT=yes \
  --env NEO4JLABS_PLUGINS='["bloom"]' \
  test/19971
```
The default location (inside the docker container) for the plugin license is `${NEO4J_HOME}/licenses/`

### DIFF
--- a/docker-image-src/4.3/docker-entrypoint.sh
+++ b/docker-image-src/4.3/docker-entrypoint.sh
@@ -419,6 +419,11 @@ if [ -d /data ]; then
     fi
 fi
 
+if [ -d /licenses ]; then
+    check_mounted_folder_readable "/licenses"
+    : ${NEO4J_dbms_directories_licenses:="/licenses"}
+fi
+
 declare -A COMMUNITY
 declare -A ENTERPRISE
 
--- a/docker-image-src/4.2/docker-entrypoint.sh
+++ b/docker-image-src/4.2/docker-entrypoint.sh
@@ -419,6 +419,11 @@ if [ -d /data ]; then
     fi
 fi
 
+if [ -d /licenses ]; then
+    check_mounted_folder_readable "/licenses"
+    : ${NEO4J_dbms_directories_licenses:="/licenses"}
+fi
+
 declare -A COMMUNITY
 declare -A ENTERPRISE
 
--- a/neo4jlabs-plugins.json
+++ b/neo4jlabs-plugins.json
@@ -9,7 +9,8 @@
     "versions": "https://bloom-plugins.s3.eu-west-2.amazonaws.com/versions.json",
     "properties": {
       "dbms.unmanaged_extension_classes": "com.neo4j.bloom.server=/browser/bloom",
-      "dbms.security.procedures.unrestricted": "bloom.*"
+      "dbms.security.procedures.unrestricted": "bloom.*",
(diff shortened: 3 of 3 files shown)


docker-image-src/4.3/docker-entrypoint.sh
```diff
@@ -419,6 +419,11 @@ if [ -d /data ]; then
     fi
 fi
 
+if [ -d /licenses ]; then
+    check_mounted_folder_readable "/licenses"
+    : ${NEO4J_dbms_directories_licenses:="/licenses"}
+fi
```
jennyowen, line 425: What you have here conflicts a bit with what you wrote in the PR message.
So you say we should be able to do:
`-v $HOME/bloom.license:/var/lib/neo4j/licenses/bloom.license`

but here you're expecting the license to be at `/licenses/bloom.license`.

It's better (IMHO) to use `/licenses` as the destination for the license folder because it's shorter and less likely to be typed in wrong, so your code here is fine. If bloom is actually looking for a `$NEO4J_HOME/licenses` folder, or even to make it more compatible either way, I'd suggest putting a symlink under neo4j home to `/licenses`. 
EG `ln -s  /licenses $NEO4J_HOME/licenses`. Same for 4.2.

@eastlondoner do you know how claudio could test the bloom plugin works, or is it enough with the current plugin testing?

eastlondoner (commented):
This looks great to me.

I think that because both this and the gql plugin set `unmanaged_extension_classes` properties there might be problems if both are used. 

But that's not the fault of this PR so we should not prevent this from being merged because of that - we should look into it separately

eastlondoner:
@jennyowen currently we do not test that any of the plugins "work" because that creates an external dependency on their versions.json and the hosting fo the relevant jar files.

This is the same situation as neo4j desktop (as far as I am aware) - we only test that the mechanism works _assuming_ that a valid versions.json and jar exist on the internet.

docker-image-src/4.3/docker-entrypoint.sh
eastlondoner: @jennyowen currently we do not test that any of the plugins "work" because that creates an external dependency on their versions.json and the hosting of the relevant jar files.

This is the same situation as neo4j desktop (as far as I am aware) - we only test that the mechanism works assuming that a valid versions.json and jar exist on the internet.

I also gave Cladio the same feedback on his example and I think he's updated it now (race condition)

GlacialBoy: I fix the PR description using `-v $HOME/bloom.license:/licenses/bloom.license`, actually the bloom plugin will search both in the relative path `${NEO4J_HOME}/licenses/bloom.license` and in the absolute path `licenses/bloom.license`.
I agree to keep the simpler one

jennyowen (approved):

PR #291, opened 2021-04-26
Fix CVE-2020-14803

rogervila:
Hello,

This PR fixes CVE-2020-14803 by upgrading OpenJDK to 11.0.11

More details here:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-14803

Another solution would be to rebuild all Neo4J images on DockerHub, so the latest _openjdk:11-jdk-slim_ image would be used.

Thank you.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 include make-common.mk
 
-NEO4J_BASE_IMAGE?="openjdk:11-jdk-slim"
+NEO4J_BASE_IMAGE?="openjdk:11.0.11-jdk-slim"
 
 # Use make test TESTS='<pattern>' to run specific tests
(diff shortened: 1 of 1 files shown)


jennyowen:
@rogervila  thank you for submitting a PR, but the tag `openjdk:11-jdk-slim` is already an alias for the one you're changing it to.
See: https://github.com/docker-library/docs/blob/master/openjdk/README.md#supported-tags-and-respective-dockerfile-links

If you clear your docker cache of neo4j and openjdk images it should re-download with the latest CVE fix.

jennyowen:
I should add that docker automatically rebuilds official images every time one of the base images updates, so if OpenJDK has the fix then we should too.

rogervila:
Thank you @jennyowen I will rebuild the container.

PR #302, opened 2021-07-09
Added apoc-core

renetapopova:
Trello cards:
* [Add apoc-core as a plugin to docker](https://trello.com/c/ezWQQW1z/536-add-apoc-core-as-a-plugin-to-docker)
* [Update docs about using `apoc core` with docker](https://trello.com/c/UQ2MhGpI/2926-update-docs-about-using-apoc-core-with-docker)

### DIFF
--- a/docker-image-src/4.4/docker-entrypoint.sh
+++ b/docker-image-src/4.4/docker-entrypoint.sh
@@ -146,6 +146,30 @@ function check_mounted_folder_writable_with_chown
     fi
 }
 
+function install_plugin_from_location
+{
+  # Install a plugin from location at runtime.
+  local _plugin_name="${1}" #e.g. apoc-core
+
+  local _plugins_dir="${NEO4J_HOME}/plugins"
+  if [ -d /plugins ]; then
+    local _plugins_dir="/plugins"
+  fi
+  local _location="$(jq --raw-output "with_entries( select(.key==\"${_plugin_name}\") ) | to_entries[] | .value.location" /neo4jlabs-plugins.json )"
+  # Using the same name for the plugin irrespective of version ensures we don't end up with different versions of the same plugin
+  local _destination="${_plugins_dir}/${_plugin_name}.jar"
+  local _neo4j_version="$(neo4j --version | cut -d' ' -f2)"
+
+  # Now we install the plugin that is shipped with Neo4j
+  echo "Installing Plugin '${_plugin_name}' from ${_location} to ${_destination} "
+  wget -q --timeout 300 --tries 30 --output-document="${_destination}" "${_location}"
+
+  if ! is_readable "${_destination}"; then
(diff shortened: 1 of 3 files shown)


docker-image-src/4.4/docker-entrypoint.sh
```diff
@@ -213,7 +213,12 @@ function install_neo4j_labs_plugins
   local _old_config="$(mktemp)"
   cp "${NEO4J_HOME}"/conf/neo4j.conf "${_old_config}"
   for plugin_name in $(echo "${NEO4JLABS_PLUGINS}" | jq --raw-output '.[]'); do
-    load_plugin_from_github "${plugin_name}"
+    local _location="$(jq --raw-output "with_entries( select(.key==\"${_plugin_name}\") ) | to_entries[] | .value.location" /neo4jlabs-plugins.json )"
+    if "${_location}" != "null"; then
+        install_plugin_from_location "${_location}"
```
eastlondoner, line 218: @renetapopova you'll need to implement install_plugin_from_location ;-)

eastlondoner (changes requested):

eastlondoner:
Nice try.

Install_plugin_from_location
needs to take a local location (/var/lib/neo4j/labs/*) and copy it to the plugins directory using cp

neo4jlabs-plugins.json
```diff
@@ -5,14 +5,20 @@
       "dbms.security.procedures.unrestricted": "apoc.*"
     }
   },
+  "apoc-core": {
+    "location": "/labs/apoc-core-*.jar",
```
eastlondoner, line 9: I think this should be a full path so /var/lib/neo4j/labs...

docker-image-src/4.4/docker-entrypoint.sh
```diff
@@ -146,6 +146,30 @@ function check_mounted_folder_writable_with_chown
     fi
 }
 
+function install_plugin_from_location
+{
+  # Install a plugin from location at runtime.
+  local _plugin_name="${1}" #e.g. apoc-core
```
eastlondoner, line 152: The passed argument will be the location not plug-in name

eastlondoner (changes requested):

renetapopova:
@eastlondoner I decided to keep the `plugin_name`as a parameter so that the function follows the same pattern as the rest; I can validate that the destination is readable, and the message is the same. I haven't tested it, I am also not sure what will happen if there are more apoc-core jar files in the location (in my Neo4j's` /labs` folder there are no jars), I guess the function will copy them all one by one.

eastlondoner (commented):
@renetapopova this looks great - leave the PR up and I will add a test

jennyowen (approved):

eastlondoner (approved):
LGTM please Squash and merge :-)

PR #320, opened 2021-12-01
Add open options settings for test configs

MishaDemianenko:


src/test/java/com/neo4j/docker/neo4jserver/TestMounting.java
```diff
@@ -240,6 +242,9 @@ void testCantWriteIfSecureEnabledAndNoPermissions_logs() throws IOException
 	void canMountAllTheThings_fileMounts(boolean asCurrentUser) throws Exception
 	{
 		Path testOutputFolder = HostFileSystemOperations.createTempFolder( "mount-everything-" );
+		Path confFile = Paths.get( "src", "test", "resources", "confs", "MountConf.conf" );
+		Files.copy( confFile, testOutputFolder.resolve( "neo4j.conf" ) );
+
 		try(GenericContainer container = setupBasicContainer( asCurrentUser, false ))
 		{
 			HostFileSystemOperations.createTempFolderAndMountAsVolume( container, "conf", "/conf", testOutputFolder );
```
jennyowen, line 250: in this test it creates a folder `/tmp/mount-everything-####/` then underneath on this line (L250) it creates a subfolder  `/tmp/mount-everything-####/conf` which is where your conf file needs to go. At the moment you're not mounting it in a location where docker will see the conf.

src/test/java/com/neo4j/docker/neo4jserver/TestMounting.java
```diff
@@ -268,7 +273,7 @@ void canMountAllTheThings_namedVolumes(boolean asCurrentUser) throws Exception
 		{
 			container.withCreateContainerCmdModifier(
 					(Consumer<CreateContainerCmd>) cmd -> cmd.getHostConfig().withBinds(
-							Bind.parse("conf-"+id+":/conf"),
+							// Bind.parse("conf-"+id+":/conf"),        // todo we need a set of open options, ask
```
jennyowen, line 276: why?

src/test/resources/confs/EnterpriseOnlyNotOverwritten.conf
```diff
@@ -1 +1,5 @@
-causal_clustering.transaction_advertised_address=localhost:6060
\ No newline at end of file
+causal_clustering.transaction_advertised_address=localhost:6060
+
+dbms.jvm.additional=--add-opens=java.base/java.nio=ALL-UNNAMED
```
jennyowen, line 3: why is it necessary to add these lines to all conf files?

jennyowen (changes requested):
I really have no idea what feature you're trying to test here or why it requires editing existing tests for things. If it's a new feature, surely it needs a new test?

MishaDemianenko:
@jennyowen i forced pushed some changes with some description as well. Also, there is one case with binds where I'm not sure how to adapt the test so will follow your suggestion there.

So in short those 3 module options that now part of custom configs are required to start neo4j and without them container will not gonna start. As result if we have any test with custom config or config directory mapping we need to adapt those configs to have those options. Otherwise neo will fail on startupt

PR #331, opened 2022-01-11
Add semantic version comparison for Plugin installation

GlacialBoy:
The release cadence in Bloom is different from the main DB so we would need to add in the `versions.js` entry like 
```
{
    "neo4j": "4.4.x",
    "version": "2.0.0",
    "jar": "https://bloom-plugins.s3.eu-west-2.amazonaws.com/bloom-plugin-4.x-2.0.0.jar"
 } 
```
so we are compatible to new version of the DB without breaking customer upgrade process.

In order to do so I've introduce the capability to add versions like "4.4.x" or "3.5.*" and in case the specific version is not found it will use these entries, if they are present.

What I did was a `jq` module with function that parse the version and perform comparisons, I've create also a small batch script for testing this function but I have no clear idea where to put it.

This is all the cases I listed and expected results ( `v1` is the version in file, `v2` is the actual docker image version ):
```
[
 { "v1": "4", "v2": "4" , "equal":true },
 { "v1": "4", "v2": "5" , "equal":false },
 { "v1": "5", "v2": "4" , "equal":false },
 { "v1": "", "v2": "4" , "equal":false },
 { "v1": "5", "v2": "" , "equal":false },
 { "v1": "", "v2": "" , "equal":false },
 { "v1": null, "v2": "1" , "equal":false },
 { "v1": "1", "v2": null , "equal":false }, 
 
 { "v1": "1.1", "v2": "1.1" , "equal":true },
 { "v1": "2.1", "v2": "1.1" , "equal":false },
 { "v1": "2.1", "v2": "2" , "equal":false },
 { "v1": "2.1", "v2": "2.2" , "equal":false },
 
 { "v1": "1.x", "v2": "1.1" , "equal":true },
 { "v1": "1.x", "v2": "2.1" , "equal":false },
 { "v1": "1.x", "v2": "2" , "equal":false },
 { "v1": "1.x", "v2": "1" , "equal":false },
 { "v1": "1.x", "v2": "1" , "equal":false },

 { "v1": "1.*", "v2": "1.1" , "equal":true },
 { "v1": "1.*", "v2": "2.1" , "equal":false },
 { "v1": "1.*", "v2": "2" , "equal":false },
 { "v1": "1.*", "v2": "1" , "equal":false },
 { "v1": "1.*", "v2": "1" , "equal":false },

 { "v1": "4.1.2", "v2": "4.1.2" , "equal":true },
 { "v1": "3.5.26", "v2": "1.5.0" , "equal":false },
 { "v1": "1.1.2", "v2": "5.1.2" , "equal":false },
 { "v1": "1.1.0", "v2": "1.1" , "equal":false },
 { "v1": "1.1.0", "v2": "1" , "equal":false },
 { "v1": "1.1.0", "v2": "" , "equal":false },

 { "v1": "3.1.x", "v2": "3.1.50" , "equal":true },
 { "v1": "3.1.x", "v2": "3.2.50" , "equal":false },
 { "v1": "3.1.x", "v2": "4.1.50" , "equal":false },
 { "v1": "3.1.x", "v2": "3.1" , "equal":false },
 { "v1": "3.1.x", "v2": "3" , "equal":false },

 { "v1": "3.1.*", "v2": "3.1.50" , "equal":true },
 { "v1": "3.1.*", "v2": "3.2.50" , "equal":false },
 { "v1": "3.1.*", "v2": "4.1.50" , "equal":false },
 { "v1": "3.1.*", "v2": "3.1" , "equal":false },
 { "v1": "3.1.*", "v2": "3" , "equal":false },

 { "v1": "3.x.x", "v2": "3.1.2" , "equal":true },
 { "v1": "3.x.x", "v2": "4.1.2" , "equal":false },
 { "v1": "3.x.x", "v2": "3.1" , "equal":false },
 { "v1": "3.x.x", "v2": "3" , "equal":false }
]
``` 

A way to test the function that is performing the comparison is saving the json above in a file `semver_data.json` and then call this function
```
function test_semver
{
  local data="$(cat ./semver_data.json)"
  local failed_tests="$(echo "${data}" | jq --raw-output "import \"semver\" as lib; .[] | select(.v2 as \$v2|.equal as \$result|.v1|lib::semver(\$v2) != \$result) ")"
  echo >&2 "Failed Tests ${failed_tests}"
}
```

I have not a clear idea how to test it inside the docker image @jennyowen so I need your support for this


GlacialBoy:
I've created a Bloom version file which contains also a 4.4.x entry here, 
https://bloom-plugins.s3.eu-west-2.amazonaws.com/versions2.json
In order to use it locally just change the `neo4jlabs-plugins.json` of Bloom with this url and start the docker image using

To install the plugin with Bloom
```
docker run -it --rm \
  --env NEO4J_AUTH=neo4j/test \
  --env NEO4J_ACCEPT_LICENSE_AGREEMENT=yes \
  --env NEO4JLABS_PLUGINS='["bloom"]' \
  neo4j:4.4.2-enterprise
```

GlacialBoy:
[Docker builds: Investigate adding wildcard support to plugin mapping JSON](https://trello.com/c/jrcx0Jw4/3049-docker-builds-investigate-adding-wildcard-support-to-plugin-mapping-json)

PR #340, opened 2022-03-22
Change to URLs for the GDS distribution

knutwalker:
In order to comply with the sanctions against Russia,
we are moving the GDS artifacts to a new bucket and URL
where we can apply the appropriate rules.


jennyowen:
Looks good! Feel free to merge it when the tests pass.

PR #355, opened 2022-07-11
add back gosu due to issues caused by incomplete shutdown

jennyowen:


jennyowen:
This is a partial revert of https://github.com/neo4j/docker-neo4j/pull/350

PR #356, opened 2022-07-12
Refine startup criteria for 5.0-Enterprise containers

gfx54b:
In *5.0 Enterprise* edition all user databases are started asynchronously. Meaning when the start is
complete ("Started" is printed in the user log) and the http endpoint responds with 200 that does not
mean that the user databases and so the default database "neo4j" is available.
All tests which assume that, are likely to fail. This change modifies the readiness criteria (only
for enterprise && >=5.0) that instead of waiting for 200 on the generic http endpoint it waits for
200 for the database's availability endpoint: /db/<dbname>/cluster/available (for this the caller
needs to authenticate).


jennyowen (approved):
This is great. THANK YOU!

PR #361, opened 2022-07-27
5.x dump test uses the new dump command semantics

PetrJanouch:


MishaDemianenko (approved):

MishaDemianenko (commented):
Looks good except that will fail for old versions

PetrJanouch:
@MishaDemianenko Why? The test should be used only for 5.x images. 4.x has another test. At least to my understanding.

MishaDemianenko:
> @MishaDemianenko Why? The test should be used only for 5.x images. 4.x has another test. At least to my understanding.

not in this repo. Check for example https://github.com/neo4j/docker-neo4j/blob/master/src/test/java/com/neo4j/docker/neo4jadmin/TestBackupRestore.java and TestBackupRestore44 over there.
There are checks like `Assumptions.assumeTrue( TestSettings.NEO4J_VERSION.isAtLeastVersion( Neo4jVersion.NEO4J_VERSION_500 )` and as you can see old versions are executed on top of your PR

MishaDemianenko:
@jennyowen can correct we if i'm wrong and something changed here (but not according to tesst code) ^^^

PetrJanouch:
That is really weird. the test I fixed already uses 5.x-only command semantics. It uses `neo4j-admin database dump`, which is something 4.x will not understand.

MishaDemianenko:
hm looks like that test class `TestDumpLoad` has this thing `Assumptions.assumeTrue( TestSettings.NEO4J_VERSION.isAtLeastVersion( Neo4jVersion.NEO4J_VERSION_500 ),` as part of `beforeAll` so version protection should be covered but that i suppose and we should be good

MishaDemianenko:
> > @MishaDemianenko Why? The test should be used only for 5.x images. 4.x has another test. At least to my understanding.
> 
> not in this repo. Check for example https://github.com/neo4j/docker-neo4j/blob/master/src/test/java/com/neo4j/docker/neo4jadmin/TestBackupRestore.java and TestBackupRestore44 over there. There are checks like `Assumptions.assumeTrue( TestSettings.NEO4J_VERSION.isAtLeastVersion( Neo4jVersion.NEO4J_VERSION_500 )` and as you can see old versions are executed on top of your PR

Whole class is already under version check so we should be good

jennyowen:
Yep, that test is 5.0 for reasons you already figured out. Looks good!

PR #406, opened 2022-11-24
Docker image will check NEO4J_AUTH password is greater than 8 chars.

jennyowen:
cl: Docker image will verify password length is greater than 8 characters before attempting to start Neo4j database.


HannesSandberg (approved):
I have problem building the project locally and can not run the test. But if it is passing it looks good to me 👍

PR #417, opened 2022-12-23
[9wb0UwLI] Reassign APOC version.json files

AzuObs:


docker-image-src/common/neo4jlabs-plugins.json
```diff
@@ -1,13 +1,12 @@
...
     "properties": {
       "dbms.security.procedures.unrestricted": "apoc.*"
     }
   },
   "apoc-core": {
     "location": "/var/lib/neo4j/labs/apoc-*-core.jar",
-    "versions": "https://github.com/neo4j/apoc/blob/master/versions.json",
     "properties": {
       "dbms.security.procedures.unrestricted": "apoc.*"
     }
```
AzuObs, line 12: The "apoc-core" config option was introduced by Reneta in 2021 and doesn't seem like it was ever intended to fetch from a remote URL. It's possible that this is because this config was only ever intended to work with Neo4j images that already contain APOC (4.2.x+, 4.3.x+, 4.4.x+, 5.x+)?

docker-image-src/common/neo4jlabs-plugins.json
```diff
@@ -1,13 +1,12 @@
 {
   "apoc": {
-    "versions": "https://github.com/neo4j/apoc/blob/master/versions.json",
+    "versions": "https://neo4j-contrib.github.io/neo4j-apoc-procedures/versions.json",
     "properties": {
       "dbms.security.procedures.unrestricted": "apoc.*"
     }
```
AzuObs, line 6: This apoc key needs to fetch from neo4j-contrib because it's used by all versions of pre-5.0 Neo4j. All of those versions of Neo4j use neo4j-contrib/apoc and need to fetch the neo4j-contrib versions.json.

docker-image-src/5/neo4jlabs-plugins.json
```diff
@@ -1,7 +1,12 @@
...
-    "versions": "https://github.com/neo4j/apoc/blob/master/versions.json",
+    "properties": {
+      "dbms.security.procedures.unrestricted": "apoc.*"
+    }
+  },
+  "apoc-core": {
+    "location": "/var/lib/neo4j/labs/apoc-*-core.jar",
     "properties": {
       "dbms.security.procedures.unrestricted": "apoc.*"
     }
```
AzuObs, line 12: I'm hesitant to introduce this option, because I believe some team members might be against it and that is why the option was removed between 4.x and 5.x 

To me, the option really just acts as an alias for apoc, and I would like to suggest adding it because I think it would be nice to be have a backwards compatible between Neo4j 4.x and 5.x.

docker-image-src/5/neo4jlabs-plugins.json
```diff
@@ -1,7 +1,12 @@
 {
   "apoc": {
     "location": "/var/lib/neo4j/labs/apoc-*-core.jar",
-    "versions": "https://github.com/neo4j/apoc/blob/master/versions.json",
+    "properties": {
+      "dbms.security.procedures.unrestricted": "apoc.*"
+    }
+  },
```
AzuObs, line 7: The way the Neo4j container works, is that it will always check the location first and it will always find an APOC jar. Thus, the versions has no effect and should be removed to reduce the chance of confusing future maintainers.

docker-image-src/5/neo4jlabs-plugins.json
```diff
@@ -1,7 +1,6 @@
 {
   "apoc": {
     "location": "/var/lib/neo4j/labs/apoc-*-core.jar",
-    "versions": "https://github.com/neo4j/apoc/blob/master/versions.json",
```
AzuObs, line 4: This shouldn't really be needed, because the file is being used by Neo4j 5.x images which all bundle APOC. I think it's a little confusing for future maintainers to add this "versions" key which isn't gonna be used.

ncordon (approved):

PR #418, opened 2023-01-09
new test to make sure apoc always loads when using NEO4J_PLUGINS

jennyowen:


jennyowen:
can't run this test on un-released neo4js.

PR #435, opened 2023-04-11
Change log retention default

stefgia:
https://trello.com/c/NYEByBVT/1030-default-tx-retention-policy-can-we-make-it-2-days-to-match-documentation

### DIFF
--- a/docker-image-src/5/docker-entrypoint.sh
+++ b/docker-image-src/5/docker-entrypoint.sh
@@ -574,7 +574,7 @@ fi
 ## these should not override *any* configurations set by the user
 
 debug_msg "Setting docker specific configuration overrides"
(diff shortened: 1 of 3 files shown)


docker-image-src/5/docker-entrypoint.sh
```diff
@@ -574,7 +574,7 @@ fi
 ## these should not override *any* configurations set by the user
 
 debug_msg "Setting docker specific configuration overrides"
-add_docker_default_to_conf "db.tx_log.rotation.retention_policy" "100M size"
+add_docker_default_to_conf "db.tx_log.rotation.retention_policy" "2 days"
```
jennyowen, line 577: same comment as 4.4 entrypoint.

jennyowen (changes requested):
can you attach the PR to the card (you attached the card to the PR)

PR #508, opened 2024-08-28
Remove plugins that are no longer supported

stefgia:


jennyowen (approved):
Looks good! Did you also check whether `n10s` is still supported in 5x?

stefgia:
n10s seems supported up to 5.20
image

PR #509, opened 2024-09-02
Add support for docker secrets

stefgia:
https://trello.com/c/MgmtbjRJ/1470-support-docker-secrets

* Env vars with `_FILE` suffix point to the file containing the env var value
* Secrets override direct env var definitions
* Support for 4.4 and 5.x added


src/test/java/com/neo4j/docker/TestDockerComposeSecrets.java
```diff
@@ -0,0 +1,113 @@
...
+    public static TemporaryFolderManager temporaryFolderManager = new TemporaryFolderManager();
+
+    private DockerComposeContainer createContainer( File composeFile, Path containerRootDir, String serviceName )
+    {
+        var container = new DockerComposeContainer( composeFile );
+
+        container.withExposedService( serviceName, DEFAULT_BOLT_PORT ).withExposedService( serviceName, DEFAULT_HTTP_PORT,
+                                                                                           Wait.forHttp( "/" ).forPort( DEFAULT_HTTP_PORT )
+                                                                                               .forStatusCode( 200 ).withStartupTimeout(
+                                                                                                       Duration.ofSeconds( 300 ) ) )
```
jennyowen, line 36: this being all mushed to the right with random line breaks makes it hard to read.

src/test/java/com/neo4j/docker/TestDockerComposeSecrets.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @RegisterExtension
+    public static TemporaryFolderManager temporaryFolderManager = new TemporaryFolderManager();
+
+    private DockerComposeContainer createContainer( File composeFile, Path containerRootDir, String serviceName )
+    {
+        var container = new DockerComposeContainer( composeFile );
+
+        container.withExposedService( serviceName, DEFAULT_BOLT_PORT ).withExposedService( serviceName, DEFAULT_HTTP_PORT,
+                                                                                           Wait.forHttp( "/" ).forPort( DEFAULT_HTTP_PORT )
```
jennyowen, line 34: I'm not sure this wait strategy will work across all versions and editions we test. 
There are a bunch of canned common wait strategies you can use in the [WaitStrategies.java](https://github.com/neo4j/docker-neo4j/blob/17dd78c568d34b079f6de3837e926fa89c6aae11/src/test/java/com/neo4j/docker/utils/WaitStrategies.java#L43) utility which will be more reliable.

The wait strategy should be on the bolt port service as well. Bolt should always work. IIRC http relies on the browser being bundled.

src/test/java/com/neo4j/docker/TestDockerComposeSecrets.java
```diff
@@ -0,0 +1,113 @@
...
+                                                                                               .forStatusCode( 200 ).withStartupTimeout(
+                                                                                                       Duration.ofSeconds( 300 ) ) )
+                 .withEnv( "NEO4J_IMAGE", getenv( "NEO4J_IMAGE" ) ).withEnv( "HOST_ROOT", containerRootDir.toAbsolutePath().toString() );
```
jennyowen, line 37: Don't do 
> `getenv( "NEO4J_IMAGE" )`

I don't think it's gonna work in cases where we set the image using maven args (which is what we do in teamcity).

Since the code seems to need the image name string, you can instead do: `TestSettings.IMAGE_ID.asCanonicalNameString()`

src/test/java/com/neo4j/docker/TestDockerComposeSecrets.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private File copyDockerComposeResourceFile( Path targetDirectory, String fileName ) throws IOException
+    {
+        File compose_file = new File( targetDirectory.toString(), fileName );
+        if ( compose_file.exists() )
+        {
+            Files.delete( compose_file.toPath() );
+        }
+        Files.copy( Objects.requireNonNull( getClass().getClassLoader().getResourceAsStream( fileName ) ), Paths.get( compose_file.getPath() ) );
```
jennyowen, line 110: I think there are enough compose files that we should put them in a subfolder of `src/test/resources`
[TestPluginInstallation.java](https://github.com/neo4j/docker-neo4j/blob/17dd78c568d34b079f6de3837e926fa89c6aae11/src/test/java/com/neo4j/docker/coredb/plugins/TestPluginInstallation.java#L142) has an example of how to retrieve files in subfolders of test resources. [Configuration.java](https://github.com/neo4j/docker-neo4j/blob/17dd78c568d34b079f6de3837e926fa89c6aae11/src/test/java/com/neo4j/docker/coredb/configurations/Configuration.java#L85) has a different method of getting them.

jennyowen (changes requested):
This looks good and will be a very cool new feature!

I have some change suggestions to clean it up a bit, and I think you're going to have test failures. There are a few of test utilities you can use, that should make the tests neater and more reliable.

docker-image-src/5/coredb/docker-entrypoint.sh
```diff
@@ -424,6 +394,36 @@ if running_as_root; then
...
+    secret_file_path="${!variable_name}"
+
+    # Check if the file exists and is readable, then read its contents
+    if [[ -f "$secret_file_path" ]]; then
+      secret_value=$(<"$secret_file_path")
+    else
+      # If it's not a file error
+      echo >&2 "The secret file '$secret_file_path' does not exist or is not readable. Make sure you have correctly configured docker secrets."
+      exit 1
+    fi
```
jennyowen, line 420: nice error behaviour! Could you make a test for it :slightly_smiling_face: 
So the conditions are:
1. secrets file does not exist
2. secrets file exists but is not readable. This could be hard to reproduce though.

Also it just occurs to me, that if there is a non-secret env with a name that ends in `_FILE` then neo4j will fail to start (or at least there would be weird behaviour). For example if someone decides have a variable like `BACKUP_FILE` pointing to a huge dump file
Maybe there needs to be a check that the secret_file_path is actually under /run/secrets before we read it into the env?

jennyowen (changes requested):

docker-image-src/5/coredb/docker-entrypoint.sh
stefgia: This didn't make sense, because container won't reach that point if file doesn't exist or is not readable. Docker handles that

jennyowen: This was the original code:
```
    if [[ -f "$secret_file_path" ]]; then
      secret_value=$(<"$secret_file_path")
    else
      # If it's not a file error
      echo >&2 "The secret file '$secret_file_path' does not exist or is not readable. Make sure you have correctly configured docker secrets."
      exit 1
    fi
```

If the file does not exist, it errors nicely.
If the file is not readable, then the container will fail messily. Having a nice error message to explain the failure is a good thing. 
There's already an `is_readable` function in the entrypoint you could use. Just skipping the check altogether so that docker just fails is a worse solution than before.

jennyowen (approved):

jennyowen (approved):

PR #520, opened 2024-10-15
Support docker secrets only for neo4j env vars

stefgia:


jennyowen (approved):
This is a good change and makes the secrets feature work much more nicely. Great work!

PR #522, opened 2024-10-17
Add support for Calendar Versioned neo4j

stefgia:


docker-image-src/2024/coredb/Dockerfile-ubi8
```diff
@@ -0,0 +1,93 @@
+FROM redhat/ubi8-minimal:latest
```
jennyowen, line 1: we don't want to support ubi8 in calver. May as well delete this file.

build-docker-image.sh
```diff
@@ -48,9 +48,19 @@ function get_compatible_dockerfile_for_os_or_error
...
     local version=${1}
     local requested_os=${2}
+
     local major=$(echo "${version}" | sed -E 's/^([0-9]+)\.([0-9]+)\..*/\1/')
     local minor=$(echo "${version}" | sed -E 's/^([0-9]+)\.([0-9]+)\..*/\2/')
+
     case ${major} in
+        # Version is calver
+        2024)
+                local SUPPORTED_IMAGE_OS=("debian" "ubi9" "ubi8")
```
jennyowen, line 58: could you remove "ubi8" from here, we don't support it any more.
(If you could also remove `"ubi8"` from the `5)` section of the switch/case that would be great too :slightly_smiling_face: )

docker-image-src/2024/neo4j-admin/Dockerfile-ubi8
```diff
@@ -0,0 +1,42 @@
+FROM redhat/ubi8-minimal:latest
```
jennyowen, line 1: also should delete this file since we don't support ubi8

jennyowen (changes requested):

jennyowen (approved):

PR #534, opened 2024-11-19
Fixed broken tests to do with java 21 and calendar versioning updates in coreDB

jennyowen:


src/test/java/com/neo4j/docker/utils/Neo4jVersionTest.java
```diff
@@ -0,0 +1,193 @@
...
+                     "A release should not be older than itself" );
+        Assertions.assertFalse( version.isNewerThan( version ) ,
+                     "A release should not be newer than itself" );
+    }
+
+    @Test
+    void testIsCalVer_2024()
+    {
+        Neo4jVersion version = new Neo4jVersion(2024, 9, 0);
+        Assertions.assertTrue(version.isCalVer(), version+" should be flagged as a calendar version");
```
stefgia, line 120: I don't think we need to make `isCalVer()` public, unless we need it in places other than tests. For docker I think the important thing would be that it converts from and to string correctly and that when compared with other version we get the expected result.

src/test/java/com/neo4j/docker/utils/Neo4jVersionTest.java
```diff
@@ -0,0 +1,193 @@
...
+        Assertions.assertEquals( "-12345", version.label, "Did not build number from " + version );
+    }
+
+    @Test
+    public void testFromVersionString_calver_releaseFormat()
+    {
+        Neo4jVersion version = Neo4jVersion.fromVersionString( "2024.10.0" );
+        Assertions.assertEquals( 2024, version.major, "Did not parse major number from " + version );
+        Assertions.assertEquals( 10, version.minor, "Did not parse minor number from " + version );
+        Assertions.assertEquals( 0, version.patch, "Did not parse patch number from " + version );
```
stefgia, line 162: Minor comment: Major, minor and patch maybe don't make sense anymore? But calling them firstDigit, secondDigit, patch is ugly so 🤷 .

src/test/resources/dockersecrets/simple-container-compose-with-external-file-var.yml
```diff
@@ -5,6 +5,6 @@ services:
       - NEO4J_ACCEPT_LICENSE_AGREEMENT=yes
       - NEO4J_AUTH=neo4j/simplecontainerpassword
       - NEO4J_DEBUG=true
-      - CERTIFICATE_FILE="CECRTIFICATE CONTENTS"
+      - CERTIFICATE_FILE="CERTIFICATE CONTENTS"
```
stefgia, line 8: That was totally intended 😂

docker-image-src/2024/coredb/docker-entrypoint.sh
```diff
@@ -159,7 +159,7 @@ function load_plugin_from_url
         echo >&2 "ERROR: could not query ${_versions_json_url} for plugin compatibility information.
     This could indicate a problem with your network or this container's network settings.
     Neo4j will continue to start, but \"${_plugin_name}\" will not be loaded."
-        return
+        return 1
```
stefgia, line 162: Is this going to break if we change `1` and `0` for `true` and `false`? Maybe not for this PR, but I just noticed

src/test/java/com/neo4j/docker/utils/Neo4jVersionTest.java
jennyowen: yeah. There's no good solution since 5.x and 4.4 are expected to still use major/minor.

docker-image-src/2024/coredb/docker-entrypoint.sh
jennyowen: I was just copying all the changes that'd happened from 5x onto calver, this was one of them. I think in bash it's considered normal to use `0` and `1` for true and false right?

PR #535, opened 2024-12-02
Alter test to handle changes that always output exceptions

mnesbit:
To aid supporting neo4j exceptions will always be output to STDERR even if not -verbose. This requires altering the docker testInvalidExtendedConfFile_nonRootUser test to have a more specific filter to validate the container doesn't restart, since the original string is repeated within the exception stack trace.


mnesbit:
Not required as we will use a different solution

PR #537, opened 2024-12-11
improved build scripts to make calendar versioned neo4js

jennyowen:


neo-tobias (approved):
Not sure if series is the best name for what used to be 
major and now is year.

neo-tobias:
but I'm ok with series so go forward and iterate and improve later

PR #543, opened 2025-01-10
added generic plugin version matching, so that plugin teams can specify a catch-all plugin with "x.x.x" or "*.*.*"

jennyowen:


docker-image-src/common/semver.jq
```diff
@@ -1,6 +1,6 @@
 def _semver_obj2obj($req):
   if   . == $req  then true
-  elif .major != $req.major    then false
+  elif .major != $req.major and .major != "x" and .major != "*" then false
```
neo-tobias, line 3: A lot of ifs and ands checking x and *, hard for me to grasp the correctness of this :-) 

Maybe this from ChatGPT is easier to read? It's just a suggestion.

-----------
# Normalize input: Replace * with x
def normalize_wildcards:
  if type == "string" then gsub("\\*"; "x") else . end;

# Modified _semver_obj2obj with single wildcard (x)
def _semver_obj2obj($req):
  if   . == $req  then true
  elif .major != $req.major and .major != "x" then false
  elif .minor != $req.minor and .minor != "x" then false
  elif .patch != $req.patch and .patch != "x" then false
  elif $req.minor == null and .minor == "x" then false
  elif $req.patch == null and .patch == "x" then false
  elif $req.major == null and $req.minor == null and $req.patch == null then false
  else true end;

# Use normalized version in _ver2obj and semver
def _ver2obj:
  normalize_wildcards |
  if   type == "object" then .
  elif type == "string" and test("(?<major>[0-9x]+)(\\.(?<minor>[0-9x]+))?(\\.?(?<patch>[0-9x]+))?") then capture("(?<major>[0-9x]+)(\\.(?<minor>[0-9x]+))?(\\.?(?<patch>[0-9x]+))?")
  elif type == "string" and . == "" then {major: null, minor: null, patch: null}
  elif type == "number" then {minor: floor, patch: (.-floor)}
  else {major: .} end;

# Main semver function
def semver($req):
  if $req == null or $req == "" then false
  elif . == $req then true
  else _ver2obj | _semver_obj2obj($req | _ver2obj) end;
  
  
 ------------ 
  Key Changes:
Normalize Wildcards with normalize_wildcards:

Convert * to x for consistency across the logic.
Applied early in _ver2obj.
Simplified _semver_obj2obj:

Removed checks for * since everything is normalized to x.
Improved Readability:

The logic becomes easier to follow, with fewer conditions to check.
Summary:
You don’t need both x and * in the logic. Standardizing on one (e.g., x) and normalizing inputs simplifies the code while maintaining full functionality.

neo-tobias (approved):
I think its good

PR #573, opened 2026-02-04
updated trixie dockerfiles to use useradd/groupadd instead of adduser/addgroup

jennyowen:
Fixes issue raised by docker here:
https://github.com/docker-library/official-images/pull/20785#issuecomment-3843101265


ericsporre (approved):
Should we not make the UBI image for calver also use JDK25?

jennyowen:
@ericsporre very good point. Fixed.

PR #589, opened 2026-07-20
Update tests to accept incubator warnings

pontusmelke:
We will soon include the `jdk.incubator.vector` by default and must update tests to accept that, and only that, warning.


jennyowen (approved):
a very neat solution!

